PR #12, opened 2017-05-12
Refactor ui a bit

laurieboyes:
A few early steps towards the end goal of 'refactoring the massive out of this', including:

- Make all references to a `myFtFeature`/`relationship` (e.g. followed) consistently `relationship`, as this makes the most sense I think
- Extract these maps of variables for each relationship into their own files so that they can be shared between smaller, broken-up files

This is a forerunner to the CAPI v2 compatibility stuff coming in.

A bit of manual testing has been done with next-article, but I'll do much more before I release. Also looking into getting in a demo app, which will be useful for pa11y at the very least (easier said than done)

### DIFF
--- a/myft/ui/index.js
+++ b/myft/ui/index.js
@@ -17,33 +17,10 @@ let flags;
 let results = {};
 let initialised;
 
-const types = {
-	saved: 'content',
-	followed: 'concept',
-	preferred: 'preference',
-	contained: 'content'
-};
-
-const actors = {
-	saved: 'user',
-	followed: 'user',
-	preferred: 'user',
-	contained: 'list'
-};
-
-const uiSelectors = {
-	saved: '[data-myft-ui="saved"]',
-	followed: '[data-myft-ui="follow"]',
-	preferred: '[data-myft-ui="prefer"]',
-	contained: '[data-myft-ui="contained"]'
-};
-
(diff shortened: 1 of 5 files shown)


myft/ui/index.js
```diff
@@ -17,41 +17,18 @@ let flags;
...
+const typesMap = require('./relationship-maps/types');
 
 const nNotificationMsgs = {
 	followAnon: `Please <a href="${subscribeUrl}" data-trackable="Subscribe">subscribe</a> or <a href="${signInLink}" data-trackable="Sign In">sign in</a> to add this topic to myFT.`,
 	saveAnon: `Please <a href="${subscribeUrl}" data-trackable="Subscribe">subscribe</a> or <a href="${signInLink}" data-trackable="Sign In">sign in</a> to save this article.`,
 	opted: 'You‘ve opted into our new site. You can return to FT.com at any time.'
 };
 
-function myFtFeatureFromEvent (ev) {
+function getRelationshipFromEvent (ev) {
```
keirog: This is my favourite bit

laurieboyes: Because of the romance?

keirog: Because it was near the top and I hadn’t finished reviewing 😄

PR #17, opened 2017-05-23
Chunky refactor of clientside myFT button press handling with lots of tests

laurieboyes:
Prelude to this work-in-progress which will mean we can send CAPIv2 `directType` instead of CAPIv1 taxonomy: https://github.com/Financial-Times/n-myft-ui/compare/more-refactoring...direct-type-instead-of-taxonomy

Splits out a lot of the confusing logic in the `getInteractionHandler` function in myft-buttons.js into unit-tested modules.

There's still plenty of weirdness within, but this work at least separates it out and makes it more digestible. Weirdness includes:

- The _relationship-config.js_ file. This stores few properties that we need for each relationship type. It's pretty weird. Seems like all of this could be in data attributes on the form itself, and then perhaps all the different relationships could work on the same generic template  ¯\\\_(ツ)\_/¯
- The collection add/remove code still feels like a hack of the other button stuff. But anyway, I've done a bit to separate its logic into its own file. Feels like ideally it would have it's own separate code path.
- Collections does as many user-followed->concept requests as there are concepts within it, whereas really it should be doing this in a single request. This would mean tweaking the myFT client to allow access to the many-at-a-time stuff
- Probably more...


myft/ui/myft-buttons/init.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export default function (opts) {
+	if (initialised) {
+		return;
+	} else {
+		initialised = true;
+
+		if (opts && opts.anonymous) {
+			anonEventListeners()
```
debugwand: semi-colon

laurieboyes: Another PR coming very soon 😁

PR #19, opened 2017-05-24
Accept directType instead of taxonomy

laurieboyes:
### Now directType compatible:
- follow button
- unfollow button
- instant alert button
- collections
- digest promo

### To do:
- [ ] make absolutely sure the react buttons aren’t used anywhere (they still just look at taxonomy)
- [ ] delete all the react button code


myft/ui/myft-buttons/collections.js
```diff
@@ -7,16 +7,43 @@ const idProperty = relationshipConfigs['followed'].idProperty;
...
-	}));
+	// CAPI2_CLEANUP
+	let taxonomies = [];
+	if (rawFormData.taxonomy) {
+		taxonomies = rawFormData.taxonomy.split(',');
+	}
+
+	let directTypes = [];
+	if (rawFormData.directType) {
+		directTypes = rawFormData.directType.split(',');
```
debugwand: this and the one above could be ternarys

PR #21, opened 2017-05-24
Remove the react versions of all buttons

laurieboyes:
With these gone, myFT UI is 100% cool with CAPIv2 `directType` instead of `taxonomy`.

_Almost completely certain_ we have nothing that still uses these. Nevertheless, I'll be good and release this as a breaking change then the worst-case scenario if I'm wrong will be that someone tries to update their stuff for CAPIv2 and they find the react buttons missing, so it'll be obviously broken and I can put them back in on a button-by-button basis in a minor release.

Also there were a load of npm dependencies we definitely don't use 🤷‍♂️. A load of the dev dependencies probably aren't used either but meh.


leggsimon:
You've removed all the dependancies. Is that right??

laurieboyes:
Ha, yeah I think so. As the only thing this component is ever used for on the server-side is to get the templates, it doesn't actually have any node dependencies. Unless you could the fact that the templates are in handlebars, but any app depending on this wouldn't have got this far without handlebars anyway 🤷‍♂️

PR #23, opened 2017-06-05
Collections: name becomes preflabel

debugwand:
* wait for myft-page and product-tour to update


debugwand:
it is a breaking change, so will be a major

keirog:
Ah, let’s catch up together about versioning before this release because I’ve claimed v6 already (in beta at the moment, and likely to stay in beta for at least another couple of weeks)

debugwand:
a couple of weeks is definitely too long, could you use a named branch instead?

PR #34, opened 2017-06-29
V7 o-migration

gvonkoss:


package.json
```diff
@@ -5,8 +5,9 @@
   "main": "server.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
+    "commit": "commit-wizard",
```
pixelandpage: Pretty sure we don't need the commit-wizard line

PR #35, opened 2017-07-04
Digest banner overlap and color fix

Thurst-ft:
Fixes for https://jira.ft.com/browse/NFT-871 and https://jira.ft.com/browse/NFT-872

Could someone confirm the changes I've made are inline with n-myft-ui styles.


GlynnPhillips:
Weirdly I can't remember what it looked like before. @Thurst-ft could you post a screen grab of how it looks after these changes so I can see what that background now looks like?

PR #36, opened 2017-07-07
changes inverse icon on hover

gvonkoss:


myft-common/main.scss
```diff
@@ -75,6 +75,12 @@
 
 		.n-myft-ui__button--inverse {
 			@include buttonWithIcon(26, 24, getColor('white'), 'plus', 'tick', getColor('black'));
+
+			&:hover {
+				&[aria-pressed="true"]:before {
```
keirog: Would you mind reducing the nesting here and just having `&:hover[aria-pressed="true"]:before`?

keirog: Or even better, `[aria-pressed="true"]:hover:before`

gvonkoss: merpo, I was trying `:before:hover` to no avail :D

PR #38, opened 2017-07-12
Save articles longer copy

pixelandpage:
Relater PR: https://github.com/Financial-Times/next-stream-page/pull/1876

For A/B testing.

Adds longer button text on myFT Save buttons on desktop devices while the myFtSaveButtonLongerCopy 'variant' flag is on for article and video pages.

Ideas on how to improve the nested handlebars logic would be welcome, but a lot of this complexity will be removed once we have the results of the A/B test (and if we're going ahead we will need a proper solution for handling Streams pages)

Gifs:
save-long-copy-article

save-long-copy-video

AND an updated Demo Page:
screen shot 2017-07-11 at 15 45 38

### DIFF
--- a/myft-common/index.js
+++ b/myft-common/index.js
@@ -13,10 +13,10 @@ module.exports = {
 
 		if (text) {
 			const textVariant = btn.getAttribute('data-text-variant');
-			const textEl = textVariant ? btn.querySelector(`.${textVariant}`) : btn;
+			const textEl = textVariant ? btn.querySelector('[data-variant-label]') : btn;
 			const alternateText = btn.getAttribute('data-alternate-text') || alternateAriaLabel;
 			textEl.textContent = alternateText;
-			if(btn.querySelector('.save-button-longer-copy')) {
+			if(textVariant) {
 				const setAltText = alternateText.includes('Saved') ? 'Save ' : 'Saved ';
 				btn.setAttribute('data-alternate-text', setAltText);
 			} else {
--- a/myft-common/index.js
+++ b/myft-common/index.js
@@ -12,8 +12,10 @@ module.exports = {
 		const text = btn.textContent || btn.innerText;
 
 		if (text) {
+			const textVariantId = btn.getAttribute('data-text-variant');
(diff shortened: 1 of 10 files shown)


myft-common/index.js
```diff
@@ -12,9 +12,16 @@ module.exports = {
 		const text = btn.textContent || btn.innerText;
 
 		if (text) {
+			const textVariant = btn.getAttribute('data-text-variant');
+			const textEl = textVariant ? btn.querySelector(`.${textVariant}`) : btn;
 			const alternateText = btn.getAttribute('data-alternate-text') || alternateAriaLabel;
-			btn.textContent = alternateText;
-			btn.setAttribute('data-alternate-text', text);
+			textEl.textContent = alternateText;
+			if(btn.querySelector('.save-button-longer-copy')) {
```
GlynnPhillips: Could we use a data attribute here instead of a class?

myft/templates/save-for-later.html
```diff
@@ -8,10 +8,36 @@
...
+		data-alternate-text="Saved "
+	{{/if}}
 	data-content-id="{{contentId}}" {{! duplicated here for tracking}}
 	aria-label="{{#if title}}Save {{title}} to myFT for later{{else}}Save this article to myFT for later{{/if}}"
 	title="{{#if title}}Save {{title}} to myFT for later{{else}}Save this article to myFT for later{{/if}}"
->{{#if buttonText}}{{buttonText}}{{else}}Save{{/if}}</button>
-</form>
+	>
+	{{!-- TODO: Tidy up this nested logic once we have the results of our A/B test--}}
+		{{~#if buttonText~}}
```
GlynnPhillips: Do we really need all the `~` ?

myft-common/index.js
pixelandpage: Yup 👍

myft/templates/save-for-later.html
pixelandpage: Good point - removing!

PR #39, opened 2017-07-19
Email digest on follow for new subscriber

geek-caroline:
- [x] check that core experience url is correct (after @laurieboyes has completed api work)
- [x] introduce the enhanced experience behaviour to work in conjunction with @laurieboyes api work

- [x] separately update next article and stream pages to include required `followPlusDigestEmail=true` data attribute when they include the 'Add to myFT' buttons


keirog:
Loving this btw 👍

geek-caroline:
- [ ] look at why it appears to be doing a double fetch
- [ ] write tests if required

PR #48, opened 2017-08-07
Concept pill WIP

laurieboyes:
Move `.n-myft-topics__item` styles into new component 'concept pill'. Also make a start on the new styles for the inverse version.

foop1

How do y'all feel about:
- ~The idea of having this in the `n-myft-ui/components` folder, as opposed to its own dedicated repo or in an `n-myft-templates` repo~ (Discussed IRL decided that this is cool)
- The name 'concept pill'. Any better ideas? Chose 'pill' over 'lozenge' 'cause I think I saw Glyndalf call it that somewhere and it's shorter and easier to swallow.
- I'm aware that this should really be a major release but we know for sure that the old `n-myft-topics__item` class is only used by the alerts page I'm tempted to just do a minor and be quick 🙃 

To do:
- [ ] Inverse added-state hover
- [ ] Choose better colours for inverse focus status
- [ ] Decide what focus state does to the icon


components/concept-pill/main.scss
```diff
@@ -0,0 +1,100 @@
...
+	color: getColor('white');
+}
+
+.n-myft-concept-pill--inverse.n-myft-concept-pill--add:before {
+	@include oIconsGetIcon('plus', getColor('white'), 10);
+	background-size: 30px;
+	margin-right: 5px;
+}
+
+.n-myft-concept-pill--inverse.n-myft-concept-pill--add.n-myft-concept-pill--added {
```
keirog: Is this level of specificity needed? Seems like we shouldn’t need `--add` here

laurieboyes: Yeah good call

components/concept-pill/main.scss
```diff
@@ -0,0 +1,100 @@
...
+	display: block;
+	@include oTypographySansBold($scale: 0);
+	font-size: 16px;
+	border: 1px solid getColor('claret');
+	border-radius: 100px;
+	padding: 5px 10px;
+	margin: 8px 13px 8px 2px;
+	white-space: nowrap;
+	max-width: 200px;
+	white-space: nowrap;
```
keirog: Duplicate property here yo. Are these grouped in any particular way? I'm a big fan of A-Z sorting, partly because it makes it easier to sport dupes

components/concept-pill/main.scss
```diff
@@ -0,0 +1,100 @@
...
+		@include oIconsGetIcon('tick', getColor('black-70'), 10);
+		background-size: 22px;
+		margin-right: 5px;
+	}
+}
+
+.n-myft-concept-pill:focus {
+	outline-style: none;
+	padding: 3px 8px;
+	border: 3px solid getColor('teal-40');
```
pixelandpage: General, but I'm not sure about teal on claret for the focus state. Does it looks a bit weird? But then I don't think moving away from the default/expected behaviour would be a good idea either 🤔 ❔

components/concept-pill/main.scss
laurieboyes: 🦅👀 Good spot! Must be a copypaste error. Will alphabetise

laurieboyes:
Aha, yep 👍

keirog:
> The name 'concept pill'. Any better ideas? Chose 'pill' over 'lozenge' 'cause I think I saw Glyndalf call it that somewhere and it's shorter and easier to swallow.

I’m cool with it

> I'm aware that this should really be a major release but we know for sure that the old n-myft-topics__item class is only used by the alerts page I'm tempted to just do a minor and be quick 

Major! We have some other changes we want to introduce so they can all come along for the ride

laurieboyes:
Sure 👍  I'll start a beta branch

GlynnPhillips:
> The name 'concept pill'. Any better ideas? Chose 'pill' over 'lozenge' 'cause I think I saw Glyndalf call it that somewhere and it's shorter and easier to swallow.

I prefer 'pill' over 'lozenge' but also think the naming shouldn't relate to its appearance which is why in the previous iteration it became 'myft-topic'. But maybe that was too vague and also should have been concept instead of topic

laurieboyes:
Yeah I was agonising over this and my confirmation bias led me to accept that naming classes based on what stuff looks like might not be _too_ bad, based on reading point 3 of this blog from this person I'd never even heard of before: http://bdavidxyz.com/blog/how-to-name-css-classes/

Part of me agrees with you that `n-myft-concept` would be better but part of me thinks that's not enough info, for example, it will compete for its meaning with `n-concept`. I thought about `n-myft-concept-link` and also `n-myft-concept-button`, but then we have use cases for it being either.

Hmmmm 🤔

keirog:
That also brings up the topic of the concept of things

laurieboyes:
Gonna merge this stuff into the new beta branch as a WIP but happy to continue to engage in the component name debate

laurieboyes:
Also, thanks all for the thorough and well-considered reviews! ❤️

PR #50, opened 2017-08-09
Add disabled concept pill to demo

laurieboyes:
🐿 v2.5.13

### DIFF
--- a/demos/templates/demo.html
+++ b/demos/templates/demo.html
@@ -58,7 +58,7 @@
 		</div>
 	</div>
 
-<div class="o-grid-row">
+	<div class="o-grid-row">
 		<div data-o-grid-colspan="12">
 			<h2>Concept pill disabled</h2>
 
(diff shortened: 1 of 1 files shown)


demos/templates/demo.html
```diff
@@ -58,6 +58,19 @@
 		</div>
 	</div>
 
+<div class="o-grid-row">
+		<div data-o-grid-colspan="12">
+			<h2>Concept pill disabled</h2>
+
+			<ul class="concept-pill-collection">
+				<li>
+					<span class="n-myft-concept-pill n-myft-concept-pill--disabled" href="#">Books</span>
```
keirog: ~`span`~ `a`?

demos/templates/demo.html
```diff
@@ -58,6 +58,19 @@
 		</div>
 	</div>
 
+<div class="o-grid-row">
```
keirog: Tab

demos/templates/demo.html
GlynnPhillips: By the looks of it this is the non-clickable version so `span` is correct ;)

laurieboyes: Oh except I left `href` in by accident, whoops 🙃

PR #56, opened 2017-08-16
Update follow plus email

geek-caroline:
Adding data-tracking-messaging element


geek-caroline:
Ah, welcome to our new o-tracking idea, sending anything prefixed with `data-trackable-` will be sent along as a parameter in the tracking call.  The irony being that here you've spotted a typo by me, it should be `data-trackable-messaging=` :)  So the payload sent through by o-tracking will now include: `messaging: followPlusEmail`, then combine this with the fact there will be an event type of 'click' sent, we'll know they clicked a button that gave them this action :)

geek-caroline:
@keirog do you think it's ok now?

keirog:
Oh ha! I fell into the tracking trap too. Yep looks good

keirog:
@geek-caroline Has the new functionality been written yet? Just want to make sure the [existing data-trackable-* bits](https://github.com/Financial-Times/n-ui/blob/master/components/n-ui/tracking/ft/utils/getDomPath.js#L44) don’t get sent too

geek-caroline:
Ooooh, that's interesting, I didn't think of the fact that other stuff might already have a data-trackable-* thing being sent...

@keirog do you know if there's any others?

keirog:
@geek-caroline Not that I know of… But not even sure if those existing ones are even used tbh ¯\_(ツ)_/¯

PR #58, opened 2017-08-23
Extra tracking for follow and save

debugwand:
🐿 v2.5.14

### DIFF
--- a/myft/test/ui/lib/tracking.spec.js
+++ b/myft/test/ui/lib/tracking.spec.js
@@ -1,0 +1,130 @@
+const sinon = require('sinon');
+const expect = require('chai').expect;
+describe('Tracking', () => {
+	const followEvent = {
+		subjectType: 'concept',
+		action: 'add',
+		subjectId: '1234',
+		data: {}
+	};
+	const unfollowEvent = {
+		subjectType: 'concept',
+		action: 'remove',
+		subjectId: '4321',
+		data: {}
+	};
+	const instantOn = {
+		subjectType: 'concept',
+		action: 'update',
+		subjectId: '5678',
+		data: {
+			_rel: {
+				instant: 'true'
+			}
+		}
+	};
+	const instantOff = {
+		subjectType: 'concept',
+		action: 'update',
+		subjectId: '8765',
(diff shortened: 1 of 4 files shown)


myft/ui/lib/tracking.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * sendEvent - dispatches event to oTracking
+ * @param {Object} customData extra data to send through in the custom event. { action: 'someaction', thingtype_id: 'someid'}
+ */
+const sendEvent = (customData) => {
+	const detail = {
+		category: 'myFT'
+	};
+	Object.assign(detail, customData);
```
laurieboyes: Is it ok that someone could intentionally/accidentally overwrite `category`?

myft/test/ui/lib/tracking.spec.js
```diff
@@ -0,0 +1,130 @@
...
+	afterEach(() => {
+		document.body.dispatchEvent.restore();
+		window.CustomEvent.restore();
+	});
+	it('sends an event for followed concepts', () => {
+		tracking.custom(followEvent);
+		expect(window.CustomEvent.called).to.be.true;
+		expect(document.body.dispatchEvent.called).to.be.true;
+		expect(window.CustomEvent.getCall(0).args[0]).to.equal('oTracking.event');
+		expect(window.CustomEvent.getCall(0).args[1]).to.have.deep.property('detail.category', 'myFT');
```
laurieboyes: Interested to know what you prefer about testing the properties individually rather than doing `deep.equals()` on the whole thing against an expected object literal

debugwand: 1) the stack trace will give me an exact line and 2) deep equals on the whole object is a maintenance headache.

myft/ui/lib/tracking.js
debugwand: I can change it so that can't happen

myft/test/ui/lib/tracking.spec.js
laurieboyes: A strong case 👍

PR #60, opened 2017-09-08
Pill improvements

keirog:
Changes to allow the pills to be `<buttons>` as well as `<a>`s. Some minor tidy-ups included. There’s an existing issue with the disabled state but I’ll fix that in a separate PR.

No visual differences to show 😊

### DIFF
--- a/components/concept-pill/main.scss
+++ b/components/concept-pill/main.scss
@@ -7,8 +7,8 @@
 }
 
 @mixin plus-icon($color) {
-	content: '';
 	@include oIconsGetIcon('plus', getColor($color), 16);
+	content: '';
 	height: 14px;
 	background-size: 38px;
 	margin-right: 7px;
@@ -17,8 +17,10 @@
 
 .n-myft-concept-pill {
 	@include oTypographySansBold($scale: 0);
+	background-color: transparent;
 	border: 1px solid getColor('claret');
 	border-radius: 100px;
+	box-sizing: content-box;
 	color: getColor('claret');
 	display: block;
 	font-size: 16px;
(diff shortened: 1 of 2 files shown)


components/concept-pill/main.scss
```diff
@@ -17,37 +20,47 @@
 
 .n-myft-concept-pill {
 	@include oTypographySansBold($scale: 0);
+	background-color: transparent;
 	border: 1px solid getColor('claret');
 	border-radius: 100px;
+	box-sizing: content-box;
```
debugwand: this is default, is it overwriting something?

components/concept-pill/main.scss
```diff
@@ -17,37 +20,47 @@
...
 }
 
 .n-myft-concept-pill:hover {
 	border: 1px solid getColor('claret');
+}
 
-	&.n-myft-concept-pill--add::before {
-		@include plus-icon('black-70');
-	}
+.n-myft-concept-pill--added:hover,
```
debugwand: should these also have `:focus` for keyboard users?

components/concept-pill/main.scss
```diff
@@ -58,25 +71,33 @@
 }
 
 .n-myft-concept-pill--disabled,
-.n-myft-concept-pill--disabled:hover {
-	border: 1px solid getColor('black-70');
-	color: getColor('black-70');
+.n-myft-concept-pill[disabled],
+.n-myft-concept-pill[disabled].n-myft-concept-pill--inverse {
+	&,
+	&:hover {
```
debugwand: same as above for `:focus`

components/concept-pill/main.scss
keirog: ah it’s to make sure the same is applied for `<button>` too

components/concept-pill/main.scss
keirog: oof definitely, thanks 😳

keirog: just realised the omission was because we have a separate focus style:
focus

PR #63, opened 2017-09-21
add flash animation

tavvy:
🐿 v2.5.16

### DIFF
--- a/myft/ui/myft-buttons/main.scss
+++ b/myft/ui/myft-buttons/main.scss
@@ -1,0 +1,54 @@
+@mixin myftHeaderIconFlash {
+
+	@-webkit-keyframes myft-flash {
+		25% { opacity: 1; }
+		50% { opacity: 1; }
+	}
+
+	@keyframes myft-flash {
+		25% { opacity: 1; }
+		50% { opacity: 1; }
+	}
+
+	@-webkit-keyframes myfy-flash-white {
+		10% { -webkit-filter: brightness(0) invert(1); }
+		75% { -webkit-filter: brightness(0) invert(1); }
+	}
+
+	@keyframes myft-flash-white {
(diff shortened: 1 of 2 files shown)


myft/ui/myft-buttons/main.scss
```diff
@@ -0,0 +1,54 @@
+@mixin myftHeaderIconFlash {
+
+	@-webkit-keyframes myft-flash {
```
remybach: I think we use `autoprefixer` so you don't need to include any of these.

myft/ui/myft-buttons/main.scss
```diff
@@ -0,0 +1,54 @@
...
+
+			@include oGridRespondTo('M') {
+				width: 88px;
+			}
+
+			@include oGridRespondTo('L') {
+				width: 100px;
+			}
+
+			-webkit-animation: myft-flash linear 1.2s;
```
remybach: ... or these...

myft/ui/myft-buttons/main.scss
```diff
@@ -0,0 +1,54 @@
...
+		25% { opacity: 1; }
+		50% { opacity: 1; }
+	}
+
+	@keyframes myft-flash {
+		25% { opacity: 1; }
+		50% { opacity: 1; }
+	}
+
+	@-webkit-keyframes myfy-flash-white {
```
pixelandpage: typo? myfy --> myft

myft/ui/myft-buttons/main.scss
tavvy: nice, good catch!

tavvy: we do indeed, i verified on the built file 👍

PR #66, opened 2017-09-28
fullscreen overlays should have a higher z-index than the sticky header

pixelandpage:
🐿 v2.5.16 
To prevent the sticky header from covering the close buttons when the save articles overlay of open on XS screens.

### DIFF
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -190,6 +190,6 @@ $o-overlay-is-silent: true;
 	.o-overlay--myft-ui {
 		@include oOverlayFullscreen($fill: 'width');
 		@include oOverlayFullscreen($fill: 'height');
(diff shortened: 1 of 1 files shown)


myft/main.scss
```diff
@@ -190,5 +190,6 @@ $o-overlay-is-silent: true;
 	.o-overlay--myft-ui {
 		@include oOverlayFullscreen($fill: 'width');
 		@include oOverlayFullscreen($fill: 'height');
+		z-index: 150;
```
keirog: Could we use [`@include nUiZIndexFor('overlay')`](https://github.com/Financial-Times/n-ui-foundations/blob/master/util/_mixins.scss#L21) please?

PR #73, opened 2017-11-21
Instant alerts mini

debugwand:


myft-common/main.scss
```diff
@@ -100,7 +100,7 @@
 			@include buttonWithIcon(26, 24, getColor('teal-40'), 'mail', 'mail', getColor('white'));
 		}
 		.n-myft-ui__button--inverse {
-			@include buttonWithIcon(26, 24, getColor('teal-80'), 'mail', 'mail', getColor('white'));
+			@include buttonWithIcon(26, 24, getColor('white'), 'mail', 'mail', getColor('black'));
```
pixelandpage: Just checking; have we checked that it's safe to overwrite this style?

benbarnett: Was trying to think where else it might be used... the icon looked a little broken with teal on white 🤔

demos/fixtures/instant-alert-button.json
```diff
@@ -0,0 +1,7 @@
+{
+  "instantAlertButton": {
+    "conceptId": "00000000-0000-0000-0000-000000000000",
+    "name": "Hidden Haiku",
+		"directType": "http://www.ft.com/ontology/product/Brand"
```
pixelandpage: Whitespace knocking circle over ¯\_(ツ)_/¯

myft-common/main.scss
debugwand: also I looked further in the code and this now matches the follow button one. I think there might have been a problem with one of the previous n-ui upgrades

PR #80, opened 2018-01-09
Make the follow button look like a pill

chee:
Adds a mixins directory with a `pill-button` mixin used by both concept-pill and follow-pill.

Adds a follow-pill component that is redirected to from the current follow-pill file.

I think the standard-theme added state on the concept pill may need a little work, as it is difficult to tell apart from the non-added state (only the icon changes).

image 
image

I'd advise against merging this until every page on the internet has been checked.


chee:
This will break [cold start](https://github.com/Financial-Times/next-myft-page/blob/d68ba0e530b23b97d1eeec97281be163530046da/views/v2/cold-start.html) as it uses the `buttonClasses` variable

components/concept-pill/main.scss
```diff
@@ -1,139 +1,17 @@
...
-
-.n-myft-concept-pill--disabled,
-.n-myft-concept-pill[disabled],
-.n-myft-concept-pill[disabled].n-myft-concept-pill--inverse {
-	&,
-	&:hover {
-		border: 1px solid getColor('black-70');
-		color: getColor('black-70');
-	}
+.n-myft-concept-pill--add {
```
GlynnPhillips: Just want to check, this and other `add` selectors below in this file are only required until we update the onboarding sections (home page and feed page) to use the follow button component?

mixins/pill-button/_icons.scss
```diff
@@ -0,0 +1,16 @@
+@mixin tick-icon($color) {
```
GlynnPhillips: Could we update these to keep the mixin naming consistent, my preference would be camel cased but I don't have a strong opinion as long as they are the same

components/concept-pill/main.scss
chee: Yes, once we use `follow` in the onboarding sections `concept-pill` can describe only the plusless link-style concept button used in the emails & alerts section.

GlynnPhillips: Thanks, just wanted to make sure I was getting it all straight in my head

chee:
With the swapping of the active/inactive hierarchy, the concept links on the email digest settings page will have to be `[aria-pressed="true"]` in order to become outline rather than filled.

GlynnPhillips:
@chee we shouldn't add `aria-pressed` to pills on the the alerts page as those pills are standard anchor tags and not buttons. Aria-pressed should only be used for buttons that can be toggled.

I think we should first ask the design team if they were aware of the different state hierarchy that has been created to find out if that is what they expected, if not then we should work with them to find a solution.

chee:
@GlynnPhillips 

yes indeed, i was hoping to talk about that tomorrow.

now that the [work to support multiple themes](https://github.com/Financial-Times/n-myft-ui/pull/80/files#diff-0745b0e7056b87cb27a795b5b0adf54a) is done (which you can see in action at the bottom of the demo page, for the opinion topper oxford style), we can create a theme for the reverse case possibly and use that for the concept links. 

only realised these buttons would end up the wrong way round at the end of the day so threw that comment there to remind me it's a problem tomorrow.

chee:
## Todo before merging this

- [ ] Create a pull request in front-page for switching to the follow button on the cold start
- [ ] Create a pull request to article for switching to the opinion style on the sky article toppers
- [ ] Create a pull request to myft-page for removing the extra button classes that previously forced a concept pill style

PR #81, opened 2018-01-31
Import the button style in common rather than myft

chee:
So that it ends up in the critical css


chee:
This is how it worked before the new button styles were brought in, and it was broken before but is good now.

PR #83, opened 2018-02-05
Update pinning button styling

sourcier:
🐿 v2.5.16

### DIFF
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -196,7 +196,7 @@ $o-overlay-is-silent: true;
 	}
 }
 
-.myft-ui__pin-button {
+.myft-pin-button {
 	@include oButtons();
 	background-color: transparent;
 	border: 0;
@@ -209,7 +209,7 @@ $o-overlay-is-silent: true;
 	}
 }
 
-.myft-ui__pin-button {
+.myft-pin-button {
 	$icon-width: 36;
 	@include oButtonsIconButton($icon-name: pin, $is-icon-only: true);
 	@include oIconsGetIcon('pin', getColor('black-40'), $icon-width, $iconset-version: 1);
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -196,11 +196,35 @@ $o-overlay-is-silent: true;
 	}
 }
 
-.myft-ui__button--pin {
+.myft-ui__instant-alert-button,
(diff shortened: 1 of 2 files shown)


myft/main.scss
```diff
@@ -196,11 +196,35 @@ $o-overlay-is-silent: true;
...
+	&[aria-pressed=true],
+	&:not([disabled]):hover {
+		background-color: transparent;
+		border: 0;
+	}
+}
+
+.myft-ui__pin-button {
+	$icon-width: 28;
+	@include oIconsGetIcon('pin', #66625f, $icon-width, $iconset-version: 1);
```
benbarnett: We should be using colours from the palette rather than hex colours directly

myft/main.scss
```diff
@@ -196,11 +196,35 @@ $o-overlay-is-silent: true;
...
+	&[aria-pressed=true],
+	&:not([disabled]):hover {
+		@include oIconsGetIcon('pin', #990f3d, $icon-width, $iconset-version: 1);
+	}
+}
+
+.myft-ui__instant-alert-button {
+	$icon-width: 28;
+	@include oIconsGetIcon('mail', #66625f, $icon-width, $iconset-version: 1);
```
benbarnett: Same here re: colours

myft/templates/instant-alert.html
```diff
@@ -9,12 +9,11 @@
...
 			<input type="hidden" value="http://www.ft.com/ontology/concept/Concept" name="directType">
 		{{/if}}
+		{{#if buttonText}}{{buttonText}}{{else}}Instant alerts off{{/if}}
 		<button
 			aria-label="Get instant alerts for {{name}}"
 			aria-pressed="false"
-			class="n-myft-ui__button
-				{{#variant}} n-myft-ui__button--{{this}}{{/variant}}
-				{{#size}} n-myft-ui__button--{{this}}{{/size}}"
+			class="myft-ui__instant-alert-button"
```
benbarnett: This button is used widely across the site – are you sure we can change the class name like this? (could be fine, just worth checking)

sourcier:
screen shot 2018-02-06 at 16 40 39

sourcier:
@benbarnett screenshots added.

PR #89, opened 2018-02-12
Pin button

sourcier:


myft/ui/myft-buttons/init.js
```diff
@@ -73,6 +75,15 @@ function signedInEventListeners () {
 	});
 }
 
+const pinButtonEventListeners = () => {
+	myftApiClient.init().then(() => {
+		delegate.on('click', 'button[data-prioritise-button]', event => {
+			event.preventDefault();
+			handlePinning('SPAN' === event.target.tagName ? event.target.parentNode : event.target);
```
chee: this line looks a little fragile, relies on the view not changing something that isn't immediately obvious would be related to the js

could use `event.target.dataset.prioritiseButton` to check ?

chee:
```js
const findButton = element => element.dataset.prioritiseButton == null
  ? findButton(element.parentNode)
  : element
```

PR #92, opened 2018-02-19
Show spinner when pin button is pressed

alexandercurtis:


myft/ui/myft-buttons/init.js
```diff
@@ -80,6 +80,7 @@ const pinButtonEventListeners = () => {
 		delegate.on('click', 'button[data-prioritise-button]', event => {
 			event.preventDefault();
 			pinning(findButton(event.target));
+			setLoading(event.target);
```
sourcier: It would be goof to call setLoading() before pinning(), so that the UI will show the loading state if anything goes wrong.

myft/main.scss
```diff
@@ -226,6 +226,36 @@ $o-overlay-is-silent: true;
 	margin: 0 1px;
 }
 
+.myft-pin-button-wrapper.loading {
```
sourcier: Can we move the pin button CSS to it's own file.

alexandercurtis: Have changed code accordingly.

PR #93, opened 2018-02-19
Use new collection style (a box of lozenges)

chee:
Before:
Collection list

After:
a box of lozenges


benbarnett:
a "flamboyance" of lozenges? https://en.wikipedia.org/wiki/List_of_English_terms_of_venery,_by_animal

chee:
image

A claretting of collections

PR #103, opened 2018-03-16
add myft notification

fenglish:


package.json
```diff
@@ -78,5 +78,8 @@
     "semver": "^5.3.0",
     "sinon": "^2.2.0",
     "sinon-chai": "^2.10.0"
+  },
+  "dependencies": {
+    "@financial-times/n-teaser": "^4.8.13"
```
fenglish: @benbarnett should I add n-teaser in bower?

benbarnett: just package.json should be fine 👍

myft-notification/main.scss
```diff
@@ -0,0 +1,25 @@
+.o-header__top-link--myft__container {
+	.o-tooltip-content {
```
benbarnett: I wonder if we can use more targetted names for o-tooltip here, i.e. `myft-notification__tooltip` or something. Just in case other tooltips end up in the header we won't mess around with them?

myft-notification/index.js
```diff
@@ -0,0 +1,94 @@
...
+	return fetch(url, option)
+		.then(fetchJson)
+		.then(({ data = {} } = {}) => data.user.digest)
+		.then(data => {
+
+			if (hasUserDismissedNotification(data)) {
+				return;
+			};
+
+			const myFtIconContainer = document.querySelector('.o-header__top-link--myft__container');
```
benbarnett: Could we move the lines that add the dot to the dom into a function called something like `insertMyFtNotification`? Just makes this function a bit more self contained.

fenglish:
@benbarnett Thank you for your reviews! I have reflected all your reviews. Could you check this PR again please?

PR #104, opened 2018-03-19
Sends a tracking event for acting on pinning

GlynnPhillips:
This will allow us to hide the pinning promo if the user has already
pinned a topic.

### DIFF
--- a/myft-notification/index.js
+++ b/myft-notification/index.js
@@ -30,7 +30,7 @@ query MyFT($uuid: String!) {
 const insertMyFtNotification = (myFtIconContainer, tooltipTarget, tooltipEl) => {
 	setNotificationDot(myFtIconContainer, tooltipTarget);
 	setTooltipElementDiv(myFtIconContainer, tooltipEl);
-}
+};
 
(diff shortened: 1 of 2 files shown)


myft-notification/index.js
```diff
@@ -30,7 +30,7 @@ query MyFT($uuid: String!) {
 const insertMyFtNotification = (myFtIconContainer, tooltipTarget, tooltipEl) => {
 	setNotificationDot(myFtIconContainer, tooltipTarget);
 	setTooltipElementDiv(myFtIconContainer, tooltipEl);
-}
+};
```
chee: 😮

myft/ui/myft-buttons/pin-button.js
```diff
@@ -1,4 +1,15 @@
...
+
+const trackPinningAction = ({action}) =>
+	new CustomEvent('oTracking.event', {
+		detail: {
+			category: 'component',
+			messaging: trackingMessageId,
+			action
+		},
+		bubbles: true
+	});
```
chee: would be nice if this was in a library

PR #110, opened 2018-04-06
Myft notification: apply new design

fenglish:
- apply new design
- synchronise expansion between header and stickyheader

screen shot 2018-04-06 at 15 11 06


myft-notification/index.js
```diff
@@ -19,42 +18,89 @@ query MyFT($uuid: String!) {
...
+	if (withDot) {
+		div.classList.add('myft-notification__icon--with-dot');
+	}
+	notification.iconContainer.appendChild(div);
 };
 
-const setTooltipElementDiv = (container, tooltipEl) => {
+const setNotificationContent = (notification, data, flags) => {
+	const publishedDate = new Date(Date.parse(data.publishedDate));
+	const PublishedDateFormatted = `${publishedDate.getDate()}/${publishedDate.getMonth()+1}/${publishedDate.getFullYear()}`;
```
benbarnett: Could we use the o-date formatting function for this maybe? https://github.com/Financial-Times/o-date/blob/master/main.js#L165

myft-notification/index.js
```diff
@@ -19,42 +18,89 @@ query MyFT($uuid: String!) {
...
+	if (withDot) {
+		div.classList.add('myft-notification__icon--with-dot');
+	}
+	notification.iconContainer.appendChild(div);
 };
 
-const setTooltipElementDiv = (container, tooltipEl) => {
+const setNotificationContent = (notification, data, flags) => {
+	const publishedDate = new Date(Date.parse(data.publishedDate));
+	const PublishedDateFormatted = oDate.format(publishedDate, 'd/M/yyyy');
```
dan-searle: this should start with a lowercase "p"?

myft-notification/index.js
```diff
@@ -19,42 +18,89 @@ query MyFT($uuid: String!) {
...
+	});
+	notification.iconContainer.querySelector('.myft-notification__icon').addEventListener('keypress', (e) => {
+		toggleNotificationByKeypress(e, notification);
+	});
+
+	// for collapse icon
+	notification.contentContainer.querySelector('.myft-notification__collapse').addEventListener('click', () => {
+		notification.contentContainer.querySelector('.myft-notification .o-expander__toggle').click();
+	});
+	notification.contentContainer.querySelector('.myft-notification__collapse').addEventListener('keypress', (e) => {
```
dan-searle: If you change from the `.myft-notification__collapse` element from a `<span>` to a `<button>`, then you won't need to listen for the keypress event explicitly - as pressing enter on it will also emit a `click` event.

myft-notification/index.js
```diff
@@ -19,42 +18,67 @@ query MyFT($uuid: String!) {
...
 	const div = document.createElement('div');
-	div.setAttribute('class', 'o-header__top-link--myft__dot');
-	div.setAttribute('id', tooltipTarget);
-	container.appendChild(div);
+	div.setAttribute('class', 'o-expander myft-notification');
+	div.setAttribute('data-o-component', 'o-expander');
+	div.setAttribute('data-o-expander-shrink-to', 'hidden');
+	notification.container.appendChild(div);
+	notification.container.querySelector('.myft-notification').innerHTML = template({ items: data.articles, PublishedDateFormatted, flags, withDot});
+	oExpander.init(notification.container.querySelector('.myft-notification'), {
```
dan-searle: `const div` is already a reference to the the `myft-notification` dom node, so you can just use that instead of getting a new reference to it using `.querySelector` on lines 35 and 36:

```javascript 
div.innerHTML = template({ items: data.articles, PublishedDateFormatted, flags, withDot});
oExpander.init(div, {
```

myft-notification/synchronise-expansion.js
```diff
@@ -0,0 +1,36 @@
...
+
+function collapse (el) {
+	setState(el, 'collapsed');
+}
+
+function setState (el, state) {
+	const toggle = el.querySelector('.o-expander__toggle');
+	const content = el.querySelector('.o-expander__content');
+	toggle.setAttribute('aria-expanded', state === 'collapsed' ? 'false' : 'true');
+	content.setAttribute('aria-hidden', state === 'collapsed' ? 'true' : 'false');
```
dan-searle: Instead of replicating what o-expander does, we should call the `.expand` or `.collapse` methods on the `oExpander` instances. The problem to be solved then will be to not end up in an infinitely loop due to the custom events that oExpander will fire.

Alternatively, is there a way of avoiding needing two instances of the expanding digest list? Then this synchronisation will not need to happen.

myft-notification/index.js
```diff
@@ -19,42 +18,99 @@ query MyFT($uuid: String!) {
...
+		}
+	});
+};
+
+const synchroniseExpansion = (stickyHeader, stickyHeaderMyFtIconContainer, ftHeaderMyFtIconContainer) => {
+	stickyHeader.addEventListener('oHeader.Sticky', (e) => {
+		if (e.detail && e.detail.isActive) {
+			stickyHeaderMyFtIconContainer.querySelector('.myft-notification').appendChild(notificationExpander.contentEl);
+		} else {
+			ftHeaderMyFtIconContainer.querySelector('.myft-notification').appendChild(notificationExpander.contentEl);
```
dan-searle: You could create a function for moving the expander content:

```javascript
const moveExpanderAfter = (el) => el.appendChild(notificationExpander.contentEl);
```

It'll make it a bit less verbose, and a bit more expressive.

fenglish:
add tracking( PR: #109) after this PR merged

PR #113, opened 2018-04-11
add tracking events

fenglish:
🐿 v2.8.0

### DIFF
--- a/myft-notification/index.js
+++ b/myft-notification/index.js
@@ -5,6 +5,7 @@ import getUuidFromSession from './get-uuid-from-session';
 import { fragments as teaserFragments } from '@financial-times/n-teaser';
 import { json as fetchJson } from 'fetchres';
 import slimQuery from './slim-query';
+import dispatchTrackingEvent from './tracking';
 import templateExpander from './notification-expander.html';
 import templateToggleButton from './notification-toggle-button.html';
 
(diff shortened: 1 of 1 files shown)


myft-notification/index.js
```diff
@@ -38,8 +39,10 @@ const toggleExpander = (e) => {
 		notificationExpander.expand();
 		moveExpanderTo(e.path[1]);
 		deleteDot();
+		dispatchTrackingEvent.digestOpened(document);
```
dan-searle: As we should always dispatch an event when expanding or collapsing the oExpander, I think it would make sense to create two functions to encapsulate this:

```javascript
const showNotifications = () => {
  notificationExpander.expand();
  dispatchTrackingEvent.digestOpened(document);
};

const hideNotifications = () => {
  notificationExpander.collapse();
  dispatchTrackingEvent.digestClosed(document);
};
```

dan-searle:
I spotted a little optimisation we can do. 

Where we listen for the header sticky event here:
https://github.com/Financial-Times/n-myft-ui/blob/master/myft-notification/index.js#L138

We only actually need to move the expander if the expander is open. Most of the times this event is fired it won't be open.

We could change the `moveExpanderTo` function to check the expander state and only perform the move if it's open. (maybe change the name to `moveExpanderIfOpen` or something?)

Also, if we always want to remove the dot when the expander is opened, then that's something else that can go in the `showNotifications` function I mentioned above.

If all this is done, then the toggleExpander function just becomes this:

```javascript
const toggleExpander = (e) => {
	if (notificationExpander.isCollapsed()) {
		showNotifications();
	} else {
		hideNotifications();
	}
	moveExpanderTo(e.path[1]);
};
```

PR #116, opened 2018-04-13
sort digest articles by unread

fenglish:
🐿 v2.8.0

### DIFF
--- a/myft-notification/digest-data.js
+++ b/myft-notification/digest-data.js
@@ -4,6 +4,9 @@ import slimQuery from './slim-query';
 
 const notificationDismissTime = 'timeUserClickedMyftNotification';
 const myftNotificationsEnabled = 'myftNotificationsEnabled';
+const hasBeenRead = (targetArticle, readArticles) => {
+	return readArticles.find(readArticle => readArticle.id === targetArticle.id);
+}
 
 export default class DigestData {
 	constructor (uuid) {
@@ -16,6 +19,11 @@ export default class DigestData {
 
 			query MyFT($uuid: String!) {
 					user(uuid: $uuid) {
+						articlesFromReadingHistory {
+							articles {
+								...TeaserExtraLight
+							}
+						}
 						digest {
 							type
 							publishedDate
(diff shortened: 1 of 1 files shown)


myft-notification/digest-data.js
```diff
@@ -33,7 +41,19 @@ export default class DigestData {
...
-				this.data = data.user.digest;
+				const digestData = data.user.digest;
+				this.data = digestData;
+
+				// reading history for past 7 days
+				const articlesUserRead = data.user.articlesFromReadingHistory ? data.user.articlesFromReadingHistory.articles : [];
+				if (articlesUserRead.length > 0) {
+					const readArticles = [];
+					const unreadArticles = [];
+					digestData.articles.forEach(article => {
```
dan-searle: This is good, and pretty easy to understand. 👍 

The only thing I would suggest (because I'm a pedant!) is that, as processing the data is a separate concern to fetching the data, it might be worth extracting all the processing code out of this promise chain - like how you've done with the `hasBeenRead`.

This would make unit testing easier (if we wanted to add them in future), and also the sequence of actions being taken would more succinctly expressed:

```javascript
return fetch(url, options)
 	.then(fetchJson)	 			
 	.then(orderByUnreadFirst)
        .then(data => {
                this.data = data;
        };
```

myft-notification/digest-data.js
```diff
@@ -4,6 +4,9 @@ import slimQuery from './slim-query';
 
 const notificationDismissTime = 'timeUserClickedMyftNotification';
 const myftNotificationsEnabled = 'myftNotificationsEnabled';
+const hasBeenRead = (targetArticle, readArticles) => {
+	return readArticles.find(readArticle => readArticle.id === targetArticle.id);
```
dan-searle: it's really minor detail but you can remove the curly brackets and `return` around line 8.

benbarnett: Also very very minor from me, but would [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) be marginally faster here?

dan-searle: I would've thought so, but it seems not!

https://jsperf.com/array-find-vs-array-some

benbarnett: Well I didn't expect that!

fenglish: Interesting! Thank you for both of you!

dan-searle: Spanish Inquisition

PR #121, opened 2018-04-17
myFT notification: check whether digest data exists in response from next-api

fenglish:
🐿 v2.8.0


myft-notification/digest-data.js
```diff
@@ -7,7 +7,14 @@ const myftNotificationsEnabled = 'myftNotificationsEnabled';
 
 const hasBeenRead = (targetArticle, readArticles) => readArticles.find(readArticle => readArticle.id === targetArticle.id);
 
-const orderByUnreadFirst = ({ data = {} } = {}) => {
+const checkDigestDataExist = ({ data = {} } = {}) => {
+	if (!data || !data.user || !data.user.digest || !data.user.digest.articles) {
```
benbarnett: Should we check for `articles.length` here too?

PR #123, opened 2018-04-17
Myft notification tooltip

dan-searle:
Still need to fix tooltip positioning problems:

screen shot 2018-04-17 at 15 51 13
screen shot 2018-04-17 at 15 50 58


myft-notification/index.js
```diff
@@ -4,6 +4,7 @@ import oDate from 'o-date';
 import getUuidFromSession from './get-uuid-from-session';
 import Feedback from '../components/feedback';
 import DigestData from './digest-data';
+import NotificationTooltip from './tooltip';
```
benbarnett: Could we call this `NotificationProductAnnouncer` or something to obviously separate it from the dropdown? What do you think?

benbarnett: Just thinking the word 'notification' can easily be interpreted differently

dan-searle: Makes sense to me!

PR #126, opened 2018-04-20
Myft notification: extract common use functions

fenglish:


myft-notification/digest-data.js
```diff
@@ -1,41 +1,18 @@
...
 	// reading history for past 7 days
 	const articlesUserRead = data.user.articlesFromReadingHistory ? data.user.articlesFromReadingHistory.articles : [];
 	if (articlesUserRead.length > 0) {
-		const readArticles = [];
-		const unreadArticles = [];
-		digestData.articles.forEach(article => {
-			hasBeenRead(article, articlesUserRead) ? readArticles.push(article) : unreadArticles.push(article);
-		});
-		result.articles = unreadArticles.concat(readArticles);
+		const digestDatadivided = divideUnreadRead(digestData, articlesUserRead);
```
samuele-artuso: `digestDataDivided` would be a bit more readable.

myft-notification/digest-data.js
```diff
@@ -1,41 +1,18 @@
...
-	}
-	return data;
-};
-
-const flattenDigestSections = data => {
-	data.user.digest.articles = data.user.digest.concepts.reduce((acc, curr) => acc.articles.concat(curr.articles));
-	return data;
-};
-
 const orderByUnreadFirst = (data) => {
```
dan-searle: I think this whole `orderByUnreadFirst` function should be extracted from this file, rather than just the `divideUnreadRead` bit - they are both essential parts of the data processing concern (which does make sense to separate from the rest of this `DigestData` class.

myft-notification/digest-data.js
dan-searle: Should be `digestDataDivided` (capital D in "divided")

PR #128, opened 2018-04-23
Simplify unread/read digest sorting

dan-searle:
This reduces the code needed in `next-article` for the new ribbons.


myft-notification/fetch-digest-data.js
```diff
@@ -16,12 +16,34 @@ const checkDigestDataExist = ({ data = {} } = {}) => {
 
 const flattenDigestSections = data => {
 	data.user.digest.articles = data.user.digest.concepts.reduce((flatArticles, concept) => {
-		return flatArticles.concat(concept.articles)
+		return flatArticles.concat(concept.articles);
```
benbarnett: 🙈

myft-notification/fetch-digest-data.js
```diff
@@ -16,12 +16,34 @@ const checkDigestDataExist = ({ data = {} } = {}) => {
...
 };
 
+const decorateWithReadStatus = data => {
+	const readArticles = data.user.articlesFromReadingHistory ? data.user.articlesFromReadingHistory.articles : [];
+
+	readArticles.forEach(readArticle => {
+		const readArticleInDigest = data.user.digest.articles.find(digestArticle => digestArticle.id === readArticle.id);
+
+		if (readArticleInDigest) {
+			readArticleInDigest.read = true;
```
benbarnett: Although we’re only using this for digest experiments at the minute, I'm wondering if `read` is ambiguous with 'read' as an access permission? Perhaps `hasBeenRead`?

myft-notification/fetch-digest-data.js
```diff
@@ -16,12 +16,34 @@ const checkDigestDataExist = ({ data = {} } = {}) => {
 
 const flattenDigestSections = data => {
```
benbarnett: Not part of this PR but reading it together with the promise chain below sort of suggests this function does something else. What do you think about calling this `extractArticlesFromSections` or something?

myft-notification/fetch-digest-data.js
dan-searle: Yes, that seems better. 👍

PR #129, opened 2018-04-23
Cache digest data request

dan-searle:
The myFT notification drawer, article digest journey tail, and article digest journey ribbon, all need the same digest data, but requesting it each time is wasteful. Cache it for the page life.

This is very simplistic/MVP caching...

### DIFF
--- a/myft-notification/fetch-digest-data.js
+++ b/myft-notification/fetch-digest-data.js
@@ -43,7 +43,7 @@ const orderByUnreadFirst = data => {
 	return data;
 };
 
-export default async (uuid) => {
+const fetchData = uuid => {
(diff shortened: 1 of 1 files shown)


myft-notification/fetch-digest-data.js
```diff
@@ -86,3 +86,13 @@ export default async (uuid) => {
 		.then(decorateWithHasBeenRead)
 		.then(orderByUnreadFirst);
 };
+
+let data;
+
+export default async (uuid, force = false) => {
+	if (!data || force) {
+		data = fetchData(uuid);
```
benbarnett: do we need an `await` here?

PR #135, opened 2018-04-27
myFT Notification – louder button next to notification when unread

benbarnett:
screenshot 2018-04-27 12 15 29

### DIFF
--- a/myft-notification/index.js
+++ b/myft-notification/index.js
@@ -30,7 +30,7 @@ const toggleNotificationContent = (e) => {
 	}
 };
 
-const insertToggleButton = (targetEl, withDot, isLargeNotification, withAnimation) => {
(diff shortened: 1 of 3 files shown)


myft-notification/index.js
```diff
@@ -30,7 +30,7 @@ const toggleNotificationContent = (e) => {
 	}
 };
 
-const insertToggleButton = (targetEl, withDot, isLargeNotification, withAnimation) => {
+const insertToggleButton = (targetEl, withDot, isLargeNotification, withAnimation, digestFrequency) => {
```
dan-searle: Any more arguments than this and we should probably just pass a config object in.

benbarnett: Agree. I toyed with it for this. One more though, and its get objectified.

PR #141, opened 2018-05-10
Only sort the digest data for the notification drawer

dan-searle:
Remove the sorting by unread->read by default for the digest data. Only sort it like that for the digest notification drawer.

This fixes the problem of the unpredictable ordering of which article index you are on whilst in the article digest journey.

### DIFF
--- a/myft-notification/index.js
+++ b/myft-notification/index.js
@@ -99,6 +99,14 @@ const dismissNotification = () => {
 
 const moveNotificationContentTo = (el) => el.appendChild(notificationContentExpander.contentEl);
 
+const orderArticlesByUnreadFirst = data => {
+	data.articles.sort((a, b) => {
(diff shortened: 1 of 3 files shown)


myft-notification/index.js
```diff
@@ -99,10 +91,18 @@ const dismissNotification = () => {
...
+	});
+
+	return data;
+};
+
 let digestData;
 let notificationContentExpander;
 
-export default async (flags = {}, options = {}) => {
+export default async (flags = {}, { animate = false, enableAnnouncer = false }) => {
```
benbarnett: Do we need `{ animate = false, enableAnnouncer = false } = {}` here in case nothing is passed in?

benbarnett: e.g. 

```js
const test = ({ animate = false, hello = true }) => { console.log(animate, hello); }

test();  // throws Uncaught TypeError: Cannot destructure property `animate` of 'undefined' or 'null'.

test({}); // happy
```

PR #144, opened 2018-05-21
Add csrf tokens to follow/unfollow forms

chee:
This adds CSRF tokens to the follow/unfollow forms so we can prevent people from creating a form on their own page which causes a follow or unfollow on ft.com.

normally we'd create a separate cookie for this so that the session could be HttpOnly, and if we ever do decide to make the session HttpOnly (which is probably a wise idea, as it makes certain XSS attacks harder) then we'll want to change this out for that instead.

It will also render `@root.csrfToken` should an app which to print out the token by itself, instead of having it set by JS.

financial-times/next-myft-proxy#37 will need to be merged first or else this PR will break follow buttons altogether.

financial-times/next-myft-client#134 will need to be merged or else `DELETE` bodies will not be passed along.

### DIFF
--- a/myft/ui/lib/set-tokens.js
+++ b/myft/ui/lib/set-tokens.js
@@ -1,7 +1,7 @@
-import { $$ } from 'n-ui-foundations';
+import { $$ as findElements } from 'n-ui-foundations';
 
 export default function (token) {
-	const inputs = $$('[data-myft-csrf-token]');
+	const inputs = findElements('[data-myft-csrf-token]');
 
 	inputs.forEach(input => {
 		input.value = token;
--- a/myft/ui/lib/set-tokens.js
+++ b/myft/ui/lib/set-tokens.js
@@ -1,0 +1,9 @@
+import { $$ } from 'n-ui-foundations';
+
+export default function (token) {
+	const inputs = $$('[data-myft-csrf-token]');
+
+	inputs.forEach(input => {
+		input.value = token;
+	});
+}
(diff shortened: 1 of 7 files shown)


myft/ui/lib/set-tokens.js
```diff
@@ -0,0 +1,9 @@
+import { $$ } from 'n-ui-foundations';
```
GlynnPhillips: TIL: we have a utility named `$$`

A Panda clearing a desk with anger

myft/ui/lib/set-tokens.js
```diff
@@ -0,0 +1,9 @@
...
+
+export default function (token) {
```
GlynnPhillips: I think it would be handy to check that the token exists at somepoint and if doesn't then we should be logging that.

myft/ui/myft-buttons/init.js
```diff
@@ -94,9 +96,11 @@ export default function (opts) {
 		if (opts && opts.anonymous) {
 			anonEventListeners();
 		} else {
+			const session = Cookies.get('FTSession');
```
GlynnPhillips: I am not sure if there are any ramifications to this but could we have an OR here and use the secure token as well? I know it would mean when checking we would need to check if matches one of two things as well.

My thinking behind this is that we should obviously be moving towards the secure token and it would be nice if when we start to switch over no extra work is required here. We can then just come back and tidy up by removing the non secure token in our own time

PR #151, opened 2018-05-29
fetchDigestData returns clone of data each time it is called

dan-searle:
🐿 v2.8.0

### DIFF
--- a/myft-notification/fetch-digest-data.js
+++ b/myft-notification/fetch-digest-data.js
@@ -1,7 +1,7 @@
-const { json: fetchJson } = require('fetchres');
-const { cloneDeep } = require('lodash');
-const { fragments: teaserFragments } = require('@financial-times/n-teaser/main');
-const slimQuery = require('./slim-query');
+import { json as fetchJson } from 'fetchres';
+import { cloneDeep } from 'lodash';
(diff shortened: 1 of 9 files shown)


myft-notification/fetch-digest-data.js
```diff
@@ -1,6 +1,7 @@
-import { json as fetchJson } from 'fetchres';
-import { fragments as teaserFragments } from '@financial-times/n-teaser/main';
-import slimQuery from './slim-query';
+const { json: fetchJson } = require('fetchres');
+const { cloneDeep } = require('lodash');
+const { fragments: teaserFragments } = require('@financial-times/n-teaser/main');
+const slimQuery = require('./slim-query');
```
benbarnett: How come no more ES modules? No likey?

benbarnett: Hmm yeah, I think we're missing a build step as these need to be transpiled before that. How did that ever work 🤔

dan-searle: There just weren't tests... :-)

The existing tracking unit test and these new ones are run with a call to mocha directly in the make file: 
https://github.com/Financial-Times/n-myft-ui/blob/master/Makefile#L32

PR #154, opened 2018-05-31
myFT unread stories indicator (work in progress)

dan-searle:


components/unread-stories-indicator/get-published-since-time.js
```diff
@@ -0,0 +1,29 @@
+import { differenceInMinutes, isAfter } from 'date-fns';
+
+const SAME_VISIT_THRESHOLD = 30;
```
benbarnett: As this feels a bit of a logic puzzle (to my brain anyway!) could we call this something like `SAME_VISIT_THRESHOLD_MINUTES` to be super clear?

components/unread-stories-indicator/get-published-since-time.js
```diff
@@ -0,0 +1,29 @@
...
+	return new Date(Date.UTC(now.getFullYear(), now.getMonth(), now.getDate(), 5, 0, 0, 0)).toISOString();
+};
+
+const isValidPublishedSince = (dateToValidate, defaultPublishedSince) => {
+	return typeof dateToValidate === 'string' && isAfter(dateToValidate, defaultPublishedSince);
+};
+
+const dateIsWithinSameVisitThreshold = date => differenceInMinutes(new Date(), date) <= SAME_VISIT_THRESHOLD;
+
+export function getPublishedSinceTime (lastVisited, publishedSince) {
```
benbarnett: Obviously up to you but as we're in the `const fn = () => {}` style, is it worth making this like that too?

components/unread-stories-indicator/get-published-since-time.js
```diff
@@ -0,0 +1,29 @@
...
+	const defaultPublishedSince = getDefaultPublishedSince();
```
benbarnett: How about:

`defaultPublishedSince` => `startOfToday` or something

there's something about the word 'default' that send my mind in a different direction to what its doing

components/unread-stories-indicator/get-published-since-time.js
benbarnett: Also what do you think about naming like this:

`lastVisited` => `userLastVisitedAt`
`publishedSince` => `userLastSawPublishedSince`

components/unread-stories-indicator/test/get-published-since-time.spec.js
```diff
@@ -0,0 +1,85 @@
...
+			it('should return the default publishedSince time', () => {
+				const lastVisited = '2018-06-02T08:00:00.000Z';
+				const lastPublishedTime = getPublishedSinceTime(lastVisited, null);
+
+				expect(lastPublishedTime).to.equal('2018-06-02T05:00:00.000Z');
+			});
+		});
+	});
+
+	describe('given the user has visited today and returns after the same-visit threshold', () => {
```
benbarnett: Even though this is hardcoding a value, do you think this would be easier to understand to someone coming in fresh:

''given the user has visited today and returns at least 30 minutes later''

components/unread-stories-indicator/test/fetch-new-content.spec.js
```diff
@@ -0,0 +1,91 @@
...
+	});
+
+	afterEach(fetchMock.reset);
+
+	it('should fetch', () => {
+		expect(fetchMock.calls().matched.length).to.equal(1);
+	});
+
+	it('should return an array of articles', () => {
+		expect(Array.isArray(data)).to.equal(true);
```
benbarnett: Not essential at all but we have chai available so could use some of the helpers here. I think it gives slightly more meaningful logs if these were to fail,

e.g. `expect(data).to.be.an('array')`

components/unread-stories-indicator/test/fetch-new-content.spec.js
```diff
@@ -0,0 +1,91 @@
...
+	});
+
+	it('should return only the followed articles', () => {
+		expect(data.length).to.equal(2);
+		expect(data.some(article => article.id === READ_AND_FOLLOWED_ARTICLE)).to.equal(true);
```
benbarnett: or `expect(data).to.include({ id: READ_AND_FOLLOWED_ARTICLE })`

dan-searle: This will assert that the whole expect object is included, whereas in this case `id` is just one property of the object in `data`.

We could do this with [chai-things](https://www.npmjs.com/package/chai-things), but I'm not sure if it's worth introducing that dependency unless we were doing a lot of this sort of thing.

benbarnett: I think you can use `match` or something, but yeah maybe not worth it if it involves bringing something in

PR #155, opened 2018-06-04
Add token placeholder input to other n-myft-ui buttons

chee:
this also sets the variable used to `@root.locals.csrfToken` instead of `@root.csrfToken` so it can be added to every page controller with a middleware instead of individually to every controller
D:
that's slightly more complex but so much easier.

### DIFF
--- a/myft/templates/pin-button.html
+++ b/myft/templates/pin-button.html
@@ -2,6 +2,11 @@
 <span class="myft-pin-divider"></span>
 <div class="myft-pin-button-wrapper">
 	<form method="post" action="/__myft/api/core/prioritised/concept/{{id}}?method={{#if prioritised}}delete{{else}}put{{/if}}">
+		<input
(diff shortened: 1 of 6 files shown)


myft/templates/pin-button.html
```diff
@@ -2,6 +2,11 @@
 <span class="myft-pin-divider"></span>
 <div class="myft-pin-button-wrapper">
 	<form method="post" action="/__myft/api/core/prioritised/concept/{{id}}?method={{#if prioritised}}delete{{else}}put{{/if}}">
+		<input
```
GlynnPhillips: Could we just wrap this in an if that checks if `@root.cacheablePersonalisedUrl ` exists. This is a property that is set by myft-page to say "yes I am myFT page and you can therefor not worry about caching".

Its just an extra step so hopefully people don't accidentally copy the csrf token into an app that really shouldn't have it

PR #166, opened 2018-06-06
myft-notification 🔫 

### LINKED ISSUE #160: Move `fetch-digest-data.js` into next-article and then kill myft-notification
The vast majority of myft-notification has been removed as the experiment is now over. The only file we still need is this one: https://github.com/Financial-Times/n-myft-ui/blob/master/myft-notification/fetch-digest-data.js

This is now only included in next-article, and can therefore be moved in: https://github.com/Financial-Times/next-article/blob/master/client/components/digest-journey/main.js#L1

Once we've done that, myft-notification can be removed from this repo.

dan-searle:
Fixes #160 

 🐿 v2.8.0


benbarnett:
🥊

PR #167, opened 2018-06-08
change the storage place from localStorage to Cookie

fenglish:
🐿 v2.9.0

To get `LAST_VISITED_AT` and `NEW_ARTICLES_SINCE ` value on Server side, we change the place from localStorage to Cookie.

This change is for an implementation which displays the latest articles since the users' last visit  on myFT feed page.

### DIFF
--- a/components/unread-articles-indicator/storage.js
+++ b/components/unread-articles-indicator/storage.js
@@ -4,8 +4,17 @@ const INDICATOR_DISMISSED_AT = 'myFTIndicatorDismissedAt';
 
 const timestampToIsoDate = ts => new Date(ts).toISOString();
 
+const setCookie = (name, value) => {
+	document.cookie = `${name}=${value}`;
+};
+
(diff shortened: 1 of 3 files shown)


dan-searle:
@fenglish - talking this through with @benbarnett some more today, we've concluded that spreading the logic for this between the server and client is too problematic. This is mainly due to not knowing the user's time zone on the server side.

We're now thinking we should go back to the initial approach which was along these lines:

* server: renders all articles for followed concepts (as it currently does)
* client: solely responsible for the lastVisitedAt/newArticlesSinceTime. Will update what the server has rendered to organise the articles by newArticlesSinceTime

The server will likely hide the content by default when the feature is switched on, and then the client will show everything once it has done its work (to avoid a jarring flash of different content).

Unfortunately this means that this PR is no longer required, sorry.

fenglish:
@dan-searle No problem! Thank you for your detailed explanation!!

PR #168, opened 2018-06-11
Refresh unread articles indicator on page visibility change

dan-searle:
Unfortunately unit testing this new feature is not straightforward as `document.hidden` is not mutable. I have added a pending test though.

### DIFF
--- a/components/unread-articles-indicator/index.js
+++ b/components/unread-articles-indicator/index.js
@@ -30,7 +30,7 @@ export default () => {
 			setLastVisitedAt();
 
 			document.addEventListener('visibilitychange', () => {
-				if (!document.hidden) {
+				if (document.visibilityState === 'visible') {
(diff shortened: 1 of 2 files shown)


components/unread-articles-indicator/index.js
```diff
@@ -10,18 +10,29 @@ import {
...
+	return getUserId
+		.then(uuid => showUnreadArticlesCount(uuid, newArticlesSinceTime))
+		.then(() => {
 			setNewArticlesSinceTime(newArticlesSinceTime);
 			setLastVisitedAt();
-		})
-		.catch(() => {});
+
+			document.addEventListener('visibilitychange', () => {
+				if (!document.hidden) {
```
benbarnett: I think this has been superseded with `document.visibilityState` now.  But serves a very similar purpose: https://www.w3.org/TR/page-visibility/#dom-document-hidden

PR #169, opened 2018-06-13
Move pin-button into components

### LINKED ISSUE #163: Move `/myft/ui/myft-buttons/pin-button` into its own component
- [ ] Move `/myft/ui/myft-buttons/pin-button*` to `components/pin-button`
- [ ] Make surrogate entry points to load the files from the new location, so consumer apps don't break

dan-searle:
Fixes #163 

It seems n-topic-card is the only place outside of this repo that uses pin-button (and just the template): https://github.com/search?q=org%3AFinancial-Times+pin-button&type=Code


components/pin-button/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+	if (prioritised === 'true') {
+		myftApiClient.remove('user', null, 'prioritised', 'concept', conceptId);
+	} else {
+		myftApiClient.add('user', null, 'prioritised', 'concept', conceptId);
+	}
+
+	// These custom events are used by envoy so we know when to show or hide promos
+	document.body.dispatchEvent(trackPinningAction({
+		action: 'act'
```
dan-searle: This looks wrong. Presumably `action` should be `prioritise` or `deprioritise`? And also include the concept ID?

GlynnPhillips: I believe this is correct and that its a more generic description. eg. the user has acted/interacted with this component.

Other values might be view or close

dan-searle:
No, it looks like it's just used by https://github.com/Financial-Times/n-myft-ui/blob/master/myft/main.scss

components/pin-button/index.js
```diff
@@ -0,0 +1,47 @@
...
+		detail: {
+			category: 'component',
+			messaging: 'myft-onboarding-pinning-promo',
+			action
+		},
+		bubbles: true
+	});
+
+const findAncestor = (el, cls) => {
+	while ((el = el.parentElement) && !el.classList.contains(cls)) {}
```
chee: out of interest why have we replaced what was a neat and easier to read recursive function with a harder to read imperative while loop with an empty body that mutates a binding?

```js
const findAncestor = (element, classname) => element.classList.contains(classname)
  ? element
  : element.parentNode && findAncestor(element.parentNode)
```

also `?` re: the full word `element` being replaced with `el`

dan-searle: Yeah, I don't particularly like this function's implementation either - I had intended to revisit it, but forgot 😊. I'm happy to it change it to your nicer recursive one.

The main reason I changed the previous `setLoading` function was that it was doing two things (finding an ancestor with a particular class, and then setting a class on it). I was just trying to move to a more functional, composable kind of approach.

Re `el` or `element` - I have no strong opinion, but `el` is fairly common (at least in MDN web docs). I was mainly trying to address the previous mix of `node` and `element`.

dan-searle: I've raised a PR for to change this function #171.

chee: <3!!!!!

PR #170, opened 2018-06-13
move digest promo into components

### LINKED ISSUE #162: Move `myft-digest-promo` into components
- [x] Move contents of `myft-digest-promo`  into `components/digest-promo`
- [x] Make surrogate entry points to load the files from the new location, so consumer apps don't break

fenglish:
Fixes #162 
🐿 v2.9.0


myft-digest-promo/index.js
```diff
@@ -1,96 +1,2 @@
...
-		if(shouldShow) {
-			showPromo();
-			bindListeners();
-		}
-	});
-};
-
-module.exports = { init };
+// ! This entry point is deprecated. Use path n-myft-ui/components/digest-promo/main.
+module.exports = require('../components/digest-promo');
```
fenglish: First time, I just added `require('../components/digest-promo')` but it doesn't work. I would like to know why I need `module.exports` here? `module.exports` is in '../components/digest-promo' tho.

benbarnett: So `module.exports` is specific to each module. As this file is just a surrogate for another module,  it needs to 'export' what `require('../components/digest-promo')` has exported.

e.g

module-a.js
```js
module.exports = 'hello';
```

module-b.js
```js
const moduleA = require('module-a');

// this module now has a 'private' variable with the contents of `moduleA`, 
// but anyone `require`ing this module (B) won't have access to it unless we export it:

module.exports = moduleA;
```

fenglish: Thank you!!

PR #172, opened 2018-06-14
Add new exported function for getting the newArticlesSinceTime

dan-searle:
...which will only determine and set the new time once per page lifecycle, no matter how many times it is called.

```javascript
import { getNewArticlesSinceTime } from 'n-myft-ui/components/unread-articles-indicator';
```


fenglish:
Thank you!! 👍

PR #173, opened 2018-06-14
Use 10pm yesterday instead of 5am today as earliest new since time.

dan-searle:


components/unread-articles-indicator/chronology.js
```diff
@@ -5,7 +5,7 @@ const SAME_VISIT_THRESHOLD_MINUTES = 30;
 const getEarliestNewArticlesSince = () => {
 	const now = new Date();
 
-	return new Date(Date.UTC(now.getFullYear(), now.getMonth(), now.getDate(), 5, 0, 0, 0)).toISOString();
+	return new Date(Date.UTC(now.getFullYear(), now.getMonth(), now.getDate() - 1, 22, 0, 0, 0)).toISOString();
```
benbarnett: Will this work if today is the 1st of the month? 🤯

dan-searle: It seems to, but on reflection, I think it'll be safer to use `'date-fns/sub_days'` to do this.

fenglish:
@dan-searle 👍 date-fns has a lot of useful functions!!! I'm amazed!

PR #174, opened 2018-06-18
Add core experience for follow, save and instant buttons

maxcbc:
# Add core experience for follow, save and instant buttons

## What
- switches default method on follow, save and instant buttons to `GET`
- moves existing `POST` action on follow, save and instant buttons to `data-js-action`
- adds 'enhance-action-urls' javascript which fires on calls to `init()` which switches methods back to `POST` and switches actions back to the content of `data-js-action`
 
## Checks
🐿 v2.9.0

### DIFF
--- a/myft/ui/myft-buttons/enhance-action-urls.js
+++ b/myft/ui/myft-buttons/enhance-action-urls.js
@@ -1,0 +1,12 @@
+
+
+export default function (context = document) {
+	['follow', 'save', 'instant']
+		.forEach(type => {
+			Array.from(context.querySelectorAll(`form.n-myft-ui--${type}`))
+				.forEach(form => {
+					form.setAttribute('method', 'POST');
+					form.setAttribute('action', form.getAttribute('data-js-action'));
+				});
+		});
+}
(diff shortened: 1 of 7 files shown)


myft/ui/myft-buttons/enhance-action-urls.js
```diff
@@ -0,0 +1,12 @@
+
+
+export default function (context = document) {
+	['follow', 'save', 'instant']
+		.forEach(type => {
+			Array.from(context.querySelectorAll(`form.n-myft-ui--${type}`))
```
alexandercurtis: I don't think you need the `Array.from` in order to iterate the result of `querySelectorAll`

chee: you don't in modern browsers, though in older browsers NodeList doesn't have .forEach

i'm not sure if polyfill.io polyfills that or not yet

maxcbc: That was my thinking given that we use `Array.from(<NodeList>)` in a fair few other places

maxcbc: Judging from the polyfill.io docs isn't implemented yet.

PR #176, opened 2018-06-18
Unread articles indicator tracking

dan-searle:


components/unread-articles-indicator/tracking.js
```diff
@@ -0,0 +1,16 @@
...
+		detail,
+		bubbles: true
+	});
+
+	document.body.dispatchEvent(event);
+}
+
+export const countShown = (count, newArticlesSinceTime, cause) => dispatchEvent({
+	category: 'unread-articles-indicator',
+	action: 'update',
```
benbarnett: Can't think if `render` would work here as we've used it before for the other things. Could be convinced either way I think.

PR #179, opened 2018-06-19
Fix CSRF token template

maxcbc:
# Fix CSRF token template

## What
- fixes csrf token template as the token was namespaced to the `locals` object

## Why
- express merges `res.locals` and the provided model so `@root.locals `will not exist

## Checks
 🐿 v2.9.0

### DIFF
--- a/components/csrf-token/input.html
+++ b/components/csrf-token/input.html
@@ -1,5 +1,5 @@
 <input
 	data-myft-csrf-token
-	value="{{#if @root.cacheablePersonalisedUrl}}{{@root.locals.csrfToken}}{{/if}}"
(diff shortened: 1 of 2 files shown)


chee:
huh, i think this should be on .locals

chee:
@GlynnPhillips do things that are set on `response.locals` get added to `@root`?

i thought they got added to `@root.locals`

maxcbc:
@chee no basically the stuff in the second arg of `res.render` gets added to res.locals and then the resulting object is used to populate the template.

It means you can use it to set defaults and then override them in individual root controllers (if you want to).

GlynnPhillips:
@maxcbc @chee Yeah it gets added to the root or it also gets added to `@root._locals`

PR #185, opened 2018-06-21
Use start of today as earliest new-articles-since-time

dan-searle:


dan-searle:
The old way was actually broken, because it wasn't correctly converting from the browser's local time to UTC - it was just using 10pm yesterday UTC regardless - which would be very wrong in most time zones.

The new implementation will handle that, but the tests are broken when in a non GMT+1 timezone - e.g. Circle CI. Looking into fixing this now.

PR #192, opened 2018-07-02
Use personalised-feed-api for content source

benbarnett:
We tried next-api, but its intricate caching just can't handle this kind of operation at scale.

As personalised-feed already acts as a caching layer between the site and elasticsearch, and already filters by topics the user follows, this hopefully gives us what we need and can handle the load.

### DIFF
--- a/components/unread-articles-indicator/fetch-new-articles.js
+++ b/components/unread-articles-indicator/fetch-new-articles.js
@@ -18,9 +18,9 @@ const decorateWithHasBeenRead = (readingHistory, allArticles) => {
 	return allArticles;
 };
 
-const contentFromPeronalisedFeed = uuid => {
+const contentFromPersonalisedFeed = uuid => {
 	const url = `/__myft/api/onsite/feed/${uuid}?originatingSignals=followed&from=-24h`;
-	const options = { credentials: 'include', timeout: 5000 };
+	const options = { credentials: 'include' };
 
 	return fetch(url, options)
 		.then(fetchJson)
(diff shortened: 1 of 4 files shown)


components/unread-articles-indicator/fetch-new-articles.js
```diff
@@ -1,65 +1,60 @@
...
 			readNewArticle.hasBeenRead = true;
 		}
 	});
 
-	return data;
+	return allArticles;
 };
 
-export default function (uuid, since) {
+const contentFromPeronalisedFeed = uuid => {
```
dan-searle: Missing s in contentFromPer**s**onalisedFeed

components/unread-articles-indicator/fetch-new-articles.js
```diff
@@ -1,65 +1,60 @@
...
+	const url = `/__myft/api/onsite/feed/${uuid}?originatingSignals=followed&from=-24h`;
+	const options = { credentials: 'include', timeout: 5000 };
```
dan-searle: Does `fetch` support `timeout`? I'm curious, because I didn't think it did.

components/unread-articles-indicator/fetch-new-articles.js
```diff
@@ -1,65 +1,60 @@
...
-				annotations {
-					id
-				}
-			}
 		}
 		`;
-	const variables = { uuid, since };
+	const variables = { uuid };
 	const url = `https://next-api.ft.com/v2/query?query=${removeLineBreaks(gqlQuery)}&variables=${JSON.stringify(variables)}&source=next-myft`;
 	const options = { credentials: 'include', timeout: 5000 };
```
dan-searle: Maybe this `timeout` is redundant too?

benbarnett: Yep, good point, I think this was lifted from server-side-land

benbarnett:
@dan-searle good point, this could be DRY'd up. Maybe we should get this out there with a smaller amount of traffic and see how personalised feed handles the new load to see if this is viable first?

Issue #190, opened 2018-06-26
Make CSS entry points one per component


benbarnett:
So consumer apps import components they need by hand, rather than using a myft-common style bundle.

This way they can choose what should be critical or lazy.

benbarnett:
Need to figure out what happens when components depend on components – do apps import both stylesheets, or does the component import the styles for you?

e.g. myFT collections depend on myFT follow buttons

PR #197, opened 2018-07-18
Clean up myft-common/main

dan-searle:
A couple of small hygiene things in support of #190:

* `nUiMyftCommon` mixin param should be redundant now
* Nothing sets `$n-ui-myft-common-applied` to true, so the condition is redundant

The following repos `import "n-myft-ui/myft-common/main"`. All but two, also `@include nUiMyftCommon(false);`:

|                   | @include nUiMyftCommon(false); |
|-------------------|--------|
| next-front-page | yes   |
| next-tour-page | yes |
| next-video-page | yes |
| next-stream-page | yes |
| next-article | yes |
| next-myft-page | yes |
| next-search-page | yes |
| next-graphics-page | no |
| n-topic-card | no |

### DIFF
--- a/myft-common/main.scss
+++ b/myft-common/main.scss
@@ -25,13 +25,7 @@
 	}
 }
 
-//TODO this parameter is only necessary up to n-ui@4
-//Auto adding myft to a head stylesheet is no longer a requirement
-@mixin nUiMyftCommon ($force-head: true) {
-	@if $force-head == true {
-		@include nUiStylesheetStart('head-n-ui-core');
-	}
-
+@mixin nUiMyftCommon () {
 	.myft-ui,
 	.n-myft-ui {
 		display: inline-block;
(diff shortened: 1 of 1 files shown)


myft-common/main.scss
```diff
@@ -109,15 +103,6 @@
...
-	}
 }
 
-$n-ui-myft-common-applied: false !default;
-
-@if $n-ui-myft-common-applied == false {
-	$n-ui-myft-common-applied: true;
-	@include nUiMyftCommon;
-}
+@include nUiMyftCommon;
```
dan-searle: It seems like this should be removed from, and `next-graphics-page` and `n-topic-card` should include it themselves?

chee: yes i am on board with this. in fact, this happens in my PR too.

chee: yes

Issue #152, opened 2018-05-30
Identify list of components so we can move them 


benbarnett:
In the form of comments on this issue, lets come up with a list of components that live within n-myft-ui. If any look like they will be troublesome to move (or need more than just simply moving a file), then lets make a note of that here, too.

benbarnett:
Contrary to the description in this issue, @dan-searle, @alexandercurtis and I paired (grapefruited?) earlier and created issues for this. The components we've identified that need action so far are:

- myft-hint
- myft-digest-promo
- myft/buttons/collections
- myft/buttons/pin-button
- myft-notification

Others such as `myft-common` need a little more thought as they need more of a refactoring task.

dan-searle:
myft-notification is no more: #160

benbarnett:
myft-hint is no more: #159

fenglish:
myft-digest-promo is no more: #162

dan-searle:
myft/buttons/pin-button has been moved #169

chee:
the last thing on this list was moved in #203

PR #203, opened 2018-08-08
Move collections component js to collections directory

chee:
I'm not sure about this, it relates to this: https://github.com/Financial-Times/n-myft-ui/issues/152

I'm not sure if this is the sort of thing we should leave to the larger refactor of that directory, but I've moved it anyway and let's see what everyone thinks.

### DIFF
--- a/myft/test/ui/myft-buttons/collections.spec.js
+++ b/myft/test/ui/myft-buttons/collections.spec.js
@@ -1,6 +1,7 @@
 /* global expect sinon*/
 
 describe('Collections', () => {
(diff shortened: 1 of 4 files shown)


myft/test/ui/myft-buttons/collections.spec.js
```diff
@@ -1,6 +1,7 @@
 /* global expect sinon*/
 
 describe('Collections', () => {
+	const eventId = 'this-isnt-a-uuid';
```
alexandercurtis: i liked the irony in the previous name... but agree your change is for the best

PR #215, opened 2018-08-15
Add deploy step for Github Pages

chee:
🐿 v2.9.0
i do not know what i am doing dog

### DIFF
--- a/scripts/deploy-gh-pages.sh
+++ b/scripts/deploy-gh-pages.sh
@@ -1,0 +1,22 @@
+# set git config from circle
+git config --global user.email $GITHUB_EMAIL
+git config --global user.name $GITHUB_NAME
+
+# Add GitHub to known hosts to avoid an interactive prompt when cloning over SSH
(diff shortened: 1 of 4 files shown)


benbarnett:
🤣

scripts/deploy-gh-pages.sh
```diff
@@ -0,0 +1,22 @@
...
+ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+
+git clone $CIRCLE_REPOSITORY_URL honk --single-branch
+
+cd honk
+
+git checkout -b gh-pages
+
+mv ../bower_components .
+mv ../node_modules .
```
benbarnett: I don't know circle 2 well enough but could `mv`ing these things mess up parallel build steps?

chee: i think deploy runs after all other build steps in the same container, Bren told me so and i believed them.

PR #222, opened 2018-08-29
fixes missing mixing bug 

maxcbc:
fixes missing mixing bug arising from using nGetImage without importing n-ui-foundations

 🐿 v2.10.2

### DIFF
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -1,8 +1,4 @@
 @import 'o-buttons/main';
-@import './ui/myft-buttons/main';
-@import './ui/lists';
-@import '../components/pin-button/main';
-@import '../components/instant-alert/main';
 
 $o-forms-is-silent: false;
 @import 'o-forms/main';
@@ -14,6 +10,11 @@ $n-notification-is-silent: false;
 $o-overlay-is-silent: false;
 @import 'o-overlay/main';
 $o-overlay-is-silent: true;
+
+@import './ui/myft-buttons/main';
+@import './ui/lists';
+@import '../components/pin-button/main';
+@import '../components/instant-alert/main';
 
 .myft-ui,
 .n-myft-ui {


myft/main.scss
```diff
@@ -1,4 +1,5 @@
 @import 'o-buttons/main';
+@import "n-ui-foundations/main";
```
benbarnett: Do we need all of this? 

Can you give an example of the `nGetImage` bug?

maxcbc:
```
ERROR in ./test/main.scss
Module build failed: 
        @include nGetImage('logo', 'brand-myft', white, 16);
         ^
      No mixin named nGetImage

Backtrace:
        myft/ui/lists.scss:6
      in ~/Projects/Financial-Times/n-myft-ui/myft/ui/lists.scss (line 6, column 11)

```
Not test main.scss is just importing the prod scss. I spose we could just import n-ui-foundations/images/mixins

benbarnett: Looks like this class just isn't needed anymore. Its used by [`n-notification`](https://github.com/Financial-Times/n-notification/blob/27ebc28be18cc14981646c917c8898b104017df4/main.scss) but that imports the mixins it needs directly.

Worth just killing it from here?

benbarnett: (by this class, I mean `myft-ui__icon` https://github.com/Financial-Times/n-myft-ui/blob/master/myft/ui/lists.scss#L5)

benbarnett: I could be argued either way about where it lives. Yes `n-notification` is generic, but it does have a `type: 'myft'` option, for which it can style its messages under its own accord. 

If we follow a rule of three before putting things in shared libraries then this wouldn't currently pass that test.

dan-searle: Ah, it seems this is all more unhygienic that I thought!

It looks like **n-notification**'s `type` option can be anything - it's just used to name a modifier class:
https://github.com/Financial-Times/n-notification/blob/master/src/js/template.js#L4

I think this should either be a predefined list of types with **n-notification** owning the styles applied by the modifier class, or `type` shouldn't set a modifier class, but instead a consumer should be allowed to set a custom class on the notification element (the styles for which the consumer should own).

Currently **n-myft-ui** owns the javascript and html and is the only use of the `myft-ui__icon` class, and **n-notification** owns this styling of it.

benbarnett: > I think this should either be a predefined list of types with n-notification owning the styles applied by the modifier class

Personally I prefer this approach as it looks like the only similarity between the styles in n-notification and n-myft-ui was the classname; n-notification overrode all of it (as you said).

FWIW the myFT pages used to use this logo much more which is probably why it lives in here at the moment.

Also in the spirit of keeping n-myft-ui less presumptuous about what it reckons your page will need, importing this in `n-notification` seems to support that, too.

I suppose there is the argument though that you would never use a 'myft' n-notification without myft features on the page, and thus have n-myft-ui already. Argh. Toss a coin?

maxcbc: I think the refactoring of this is a separate issue. I'll work on this next. Solved this issue though.

> Currently n-notification/main imports n-ui-foundations/main, and this repo imports n-notification/main, so I'm not sure why SASS says there's no nGetImage mixin.

Thats the right question. Its because `n-notification/main` is imported after the local scss files. I've updated this PR to import local files after other files and have removed the explicit n-ui-foundations import

dan-searle: Nice one.

Yep, I'm working on point 2 now.

PR #224, opened 2018-08-29
Move myft class from n-notification

maxcbc:
Half of Financial-Times/n-notification#48

🐿 v2.10.2

### DIFF
--- a/myft/ui/lists.scss
+++ b/myft/ui/lists.scss
@@ -3,7 +3,21 @@
 }
 
 .myft-ui__icon {
-	@include nGetImage('logo', 'brand-myft', white, 16);
+	@include nGetImage('logo', 'brand-myft', getColor('white'), 16);
 	width: 34px;
 	margin: 0 8px -5px;
+}
+
+.n-notification {
+	.myft-ui__icon {
+		@include nGetImage('logo', 'brand-myft', getColor('black-80'), 20);
+		margin: 0 3px -5px;
+	}
+
+	&--error,
+	&--success {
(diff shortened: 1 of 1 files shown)


myft/ui/lists.scss
```diff
@@ -3,7 +3,18 @@
...
-	@include nGetImage('logo', 'brand-myft', white, 16);
+	@include nGetImage('logo', 'brand-myft', getColor('white'), 16);
 	width: 34px;
 	margin: 0 8px -5px;
+
+}
+
+.n-notification .myft-ui__icon {
+	@include nGetImage('logo', 'brand-myft', getColor('black-80'), 20);
+	margin: 0 3px -5px;
```
dan-searle: I'm pretty sure `.myft-ui__icon` is only ever used within `.n-notification`, so I think these two rules can be combined and the specificity reduced. Could also take advantage of a bit more sassiness with the modifiers:

```sass
.myft-ui__icon {
  @include nGetImage('logo', 'brand-myft', getColor('black-80'), 20);
  margin: 0 3px -5px;
  width: 34px;

  .n-notification--error &,
  .n-notification--success & {
     @include nGetImage('logo', 'brand-myft', getColor('white'), 20);
  }
}
```

maxcbc: I did think of the sassiness but I felt parent selectors made it less clear. Have updated to use `&--error, &--success` though

maxcbc: Also I'm not sure where `,myft-ui__icon` is used. I'll have a search round

PR #228, opened 2018-09-10
add save-with-icon styling for button

fenglish:
🐿 v2.10.2

screen shot 2018-09-10 at 11 32 38

### DIFF
--- a/components/button/main.scss
+++ b/components/button/main.scss
@@ -21,7 +21,7 @@
 	}
 }
 
-.n-myft-ui__button--save-with-icon {
+.n-myft-ui__save-button-with-icon {
 	$icon-size: 38px;
 
 	@include oIconsGetIcon('bookmark-outline', oColorsGetPaletteColor('black'), $icon-size, $iconset-version: 1);
@@ -34,7 +34,7 @@
 
 	span {
 		position: absolute;
-		top: calc(#{$icon-size} - 5px);
+		top: $icon-size - 5px;
 		left: 0;
 		right: 0;
 		@include oTypographySans($scale: -2);
@@ -49,9 +49,5 @@
 	&[aria-pressed=true] {
 		@include oIconsGetIcon('bookmark', oColorsGetPaletteColor('claret'), $icon-size, $iconset-version: 1);
 		background-position: 0 0;
-	}
-
-	.n-myft-ui__button--viewport-large {
-		display: none;
 	}
 }
--- a/components/button/main.scss
+++ b/components/button/main.scss
@@ -20,3 +20,38 @@
 		display: inline-block;
 	}
 }
+
+.n-myft-ui__button--save-with-icon {
+	$icon-size: 38px;
+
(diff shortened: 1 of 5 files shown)


components/button/main.scss
```diff
@@ -20,3 +20,38 @@
...
+	color: oColorsGetPaletteColor('black');
+	background-position: 0 0;
+	border: 0;
+	min-height: $icon-size;
+	width: $icon-size;
+	position: relative;
+
+	span {
+		position: absolute;
+		top: calc(#{$icon-size} - 5px);
```
benbarnett: Do we need a fallback for `calc`? Or does autoprefixer do it for us?

benbarnett:
Nice one. Can we also add a version of this to the demo pages/styleguide too?

components/button/main.scss
fenglish: @benbarnett Oh I will check that! This one is nested `calc` so maybe the support is different from normal `calc`? You mean so?

benbarnett: I think this would still output `calc` in the built CSS? If you need SASS to work it out and just output the resulting number for you, I think you can just remove `calc` and leave `#{$icon-size} - 5px` there? I might be wrong but worth checking I think

fenglish: screen shot 2018-09-10 at 10 17 37

It looks ok on IE11 🙌

benbarnett: So I think we can just use SASS to work this out for us: https://sass-lang.com/guide#topic-8

myft/templates/save-for-later.html
```diff
@@ -7,7 +7,7 @@
 	{{> n-myft-ui/components/csrf-token/input}}
 	<button
 		type="submit"
-		class="n-myft-ui__button{{#variant}} n-myft-ui__button--{{this}}{{/variant}}"
+		class="{{#ifEquals variant 'save-with-icon'}}n-myft-ui__button--save-with-icon{{else}}n-myft-ui__button n-myft-ui__button--{{variant}}{{/ifEquals}}"
```
dan-searle: I think this new Save button style should probably not be a variant of a standard `n-myft-ui__button` as it is so different.

With BEM, a _modifier_ class should always be used in conjunction with the standard _block_ class, so in this case it's a odd that the `n-myft-ui__button--save-for-later` modifier is used without `n-myft-ui__button`.

Perhaps, rather than re-using the existing `variant` option for this template, there should be a separate `textWithIcon` option that if `true` will set a different block class altogether?

Something like:

```handlebars
class="{{#if textWithIcon}}save-button-with-icon{{else}}n-myft-ui__button{{#if variant}} n-myft-ui__button--{{variant}}{{/if}}{{/if}}"
```

components/button/main.scss
fenglish: I see! SASS is clever!! Thank you for telling me!!

myft/templates/save-for-later.html
fenglish: @dan-searle I see! That makes sense. I was uncomfortable with `n-myft-ui__button` BEM problem, but I couldn't think the solution. Thank you!!

PR #229, opened 2018-09-10
Save button with icon: set alternate text into correct element

fenglish:
🐿 v2.10.2

Fix a bug which set alternate text into wrong element(directly into `button` not into `span`) after clicking save button.

```js
// before click
<button>
  <span>Save</span>
</button>
```
```js
// after click
<button>
  Saved
</button>
```
- add `data-text-variant` and `data-variant-label` which are needed to toggle the button correctly
https://github.com/Financial-Times/n-myft-ui/blob/master/myft-common/index.js#L16

### DIFF
--- a/components/button/main.scss
+++ b/components/button/main.scss
@@ -32,7 +32,7 @@
 	width: $icon-size;
 	position: relative;
 
-	span {
+	.save-button-with-icon-copy {
 		position: absolute;
 		top: $icon-size - 5px;
 		left: 0;
--- a/demos/templates/demo.html
+++ b/demos/templates/demo.html
@@ -30,10 +30,10 @@
 				{{/saveButton}}
 
 				{{#saveButton}}
-				<h2 class="demo-section__title">
-					Save button with icon
-				</h2>
-				{{> n-myft-ui/myft/templates/save-for-later saveButtonWithIcon=true }}
+					<h2 class="demo-section__title">
+						Save button with icon
+					</h2>
+					{{> n-myft-ui/myft/templates/save-for-later saveButtonWithIcon=true }}
 				{{/saveButton}}
 
 				{{#saveButton}}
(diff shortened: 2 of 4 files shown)


components/button/main.scss
```diff
@@ -32,7 +32,7 @@
 	width: $icon-size;
 	position: relative;
 
-	span {
+	.save-button-with-icon-copy {
```
benbarnett: Following BEM, we could make this a descendant of `n-myft-ui__save-button-with-icon`

However typically we don't nest more than 1 level deep, so it can sit at the 'same level' as `n-myft-ui__save-button-with-icon`.

e.g. `n-myft-ui__save-button-with-icon-copy`

Its wordy but follows the convention a bit more. What do you think?

dan-searle:
I'm starting to wonder if there's so many conditions within these save button templates, that it will make sense to have completely separate templates for the old style button and the new one?

components/button/main.scss
fenglish: @benbarnett Yes, I agree the BEM convention thing, however there is `save-button-longer-copy` class for the same situation as this, which the class is to set an alternate text into `span`.
https://github.com/Financial-Times/n-myft-ui/blob/master/myft/templates/save-for-later.html#L33

I'm not sure which one I should take, the BEM convention or consistency with the existing class 🤔

benbarnett: Ah if it's an existing element that's makes sense as you have it.

But might be worth us doing as @dan-searle suggests and eventually separating these out completely to avoid this kind of overlap. I guess we need to keep an eye out for how the javascript interacts with these classnames though, as they are controlled by the same code. One for hygiene?

fenglish:
We are not going to separate templates which @dan-searle suggested above because we will unify this into save button with icon style to keep consistency in the near future!!

PR #234, opened 2018-09-12
Move save buttons templates

alexandercurtis:
Put them with the other component templates
 🐿 v2.10.3


chee:
both of these templates are so similar. should we put an issue in to make them one thing with a option like we did with follow/unfollow?

PR #235, opened 2018-09-13
Update deploy script and CircleCI config to properly deploy docs

chee:
This will only run on master, will run parallel with master's tests because it doesn't really matter.

some of the weird stuff going on:

* added a write ssh key to the circle ci config and github to allow pushing. would be great if you could say it can only push to this one branch, but you can't. 
* it would not stop trying to use the READ-ONLY key no matter how firmly i demanded it so i am [taking away all of its keys](https://github.com/Financial-Times/n-myft-ui/blob/010e81210691c53769aa2df70629f6217a2dfc11/scripts/deploy-gh-pages.sh#L6) until it learns to play fair
* using `GIT_SSH` env variable with a wrapper script so we can tell it not to get all interactive about known hosts

--

in other news: this docs page can't get the font because of a CORS error, and for some reason the icons are missing. : ) : )


chee:
this latest green tick is good because in the latest version it doesn't run on this branch. if you look at the build history though for yesterday you can see it run and succeed in deployment. that was happening on the `autodocs` brank but is now happening on the `master` brank from now on.

chee:
(here's a successful deployment: https://circleci.com/gh/Financial-Times/n-myft-ui/1581)

scripts/deploy-gh-pages.sh
```diff
@@ -1,10 +1,13 @@
...
-# Add GitHub to known hosts to avoid an interactive prompt when cloning over SSH
-mkdir -p ~/.ssh
-ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+# hackily remove all identities, and later force the specific identity
+# otherwise git keeps using the wrong boi
+ssh-add -D
+
+git --version
+
+export GIT_SSH="`pwd`/scripts/ssh.sh"
```
alexandercurtis: i can't figure out where this is used :confused:

scripts/make-static-demo.sh
```diff
@@ -1,16 +1,12 @@
...
-}
-
 node demos/app &
 
 pid="$!"
 
 sleep 10
 
-get http://localhost:5005 public/index.html
+curl http://localhost:5005 | sed 's#/public/main.css#main.css#' | sed 's#href="/"#href="index.html"#' > public/index.html
```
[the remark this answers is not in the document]
chee: we don't

scripts/deploy-gh-pages.sh
chee: it's used in all the `git` commands

alexandercurtis: ah yes

simonplend:
@chee I noticed this PR and it reminded me of this:

- https://github.com/Financial-Times/x-dash/blob/master/.circleci/config.yml#L136-L145
- https://github.com/Financial-Times/x-dash/blob/master/private/scripts/gh-pages

Perhaps worth pulling out into a [shared helper script](https://github.com/Financial-Times/next-ci-shared-helpers)?

chee:
@simonplend i was sick back when you said this and then forgot about it.

i think it would be worth it. we have some differences that could be ironed out, but i think it's worth making this a shared helper.

for now i'll merge this in as is, but it would be great if it's something that could be really easy to add.

simonplend:
@chee :+1: Would you mind [opening a wee issue](https://github.com/Financial-Times/next-ci-shared-helpers/issues/new?title=New%20helper:%20helper-deploy-to-github-pages) that references your script in this repo and the `x-dash` one?

PR #236, opened 2018-09-19
Alex move save buttons

alexandercurtis:


myft/templates/unsave-for-later.html
```diff
@@ -1 +1 @@
-{{> n-myft-ui/components/unsave-for-later/unsave-for-later}}
+{{> n-myft-ui/components/save-for-later/save-for-later setSaveButtonStateToSelected=true }}
```
dan-searle: maybe `isSaved=true`? It's shorter and more agnostic about what will be different - e.g. it's not just the selected state, but the button text etc.

alexandercurtis: What I have is analagous to the existing code in follow buttons, which uses `setFollowButtonStateToSelected`
But I prefer your suggestion.

PR #237, opened 2018-09-19
Remove preferred relationship config

dan-searle:
It's not used. Is it?


GlynnPhillips:
I think this gets used to to personalise any forms so as long as we are sure no where is adding a form to get people to change their preferences then it can go

dan-searle:
I can't find anything that could possibly use it.

PR #239, opened 2018-09-19
relationshipConfig is only source of truth for locating button elements in UI

dan-searle:
Previously, a few places used the `n-myft-ui--{action}` css class to locate UI elements, and some places used the `uiSelector` property from the config.

This PR makes everywhere consistently use `uiSelector` config property.

The `n-myft-ui--{action}` CSS class remain only for styling purposes.

### DIFF
--- a/myft/ui/myft-buttons/enhance-action-urls.js
+++ b/myft/ui/myft-buttons/enhance-action-urls.js
@@ -1,12 +1,19 @@
+import relationshipConfig from '../lib/relationship-config';
+
 export default function (context = document) {
-	['follow', 'save', 'instant']
-		.forEach(type => {
-			Array.from(context.querySelectorAll(`form.n-myft-ui--${type}`))
+	[
+		relationshipConfig.followed.uiSelector,
+		relationshipConfig.saved.uiSelector,
+		'[data-myft-ui="instant"]'
+	]
+		.forEach(selector => {
+			Array.from(context.querySelectorAll(selector))
 				.forEach(form => {
-					const action = form.getAttribute('data-js-action');
-					if (!action) return;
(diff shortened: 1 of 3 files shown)


myft/ui/myft-buttons/enhance-action-urls.js
```diff
@@ -1,12 +1,19 @@
...
+			Array.from(context.querySelectorAll(selector))
 				.forEach(form => {
-					const action = form.getAttribute('data-js-action');
-					if (!action) return;
-					form.setAttribute('method', 'POST');
-					form.setAttribute('action', action);
+					if (form.dataset.jsAction) {
+						form.setAttribute('method', 'POST');
+						form.setAttribute('action', form.dataset.jsAction);
+						delete form.dataset.jsAction;
```
alexandercurtis: why delete?

dan-searle: It's not strictly necessary to , but if `enhanceActionUrls` is called more than once on the same form, then there's no need to set the forms `method` and `action` again, so deleting`data-js-action` is an easy way of avoid unnecessary work (and maybe unexpected behaviour - like if something else has modified the `method` or `action` in the meantime).

If we think this bit is a risky change though, then I'm happy to remove it from this PR

alexandercurtis: If you're happy with it, then I'm happy with it :smile:

I mentioned it because it was the only change to functionality in the PR.

dan-searle: yeah, well spotted, I should've made that clearer in the PR description!

PR #242, opened 2018-09-26
Separate pin button init from general myft-ui init

### LINKED ISSUE #241: Consumers should initialise pin buttons
Currently, the pin buttons are initialised as part of the general `n-myft-ui` button `init`: https://github.com/Financial-Times/n-myft-ui/blob/master/myft/ui/myft-buttons/init.js#L94

Yet, pin buttons exist as a standalone component with all the code in `n-myft-ui/components/pin-button/`

Really, whatever wants to use pin buttons ([n-topic-card](https://github.com/search?q=org%3AFinancial-Times+pin-button&type=Code)) should be responsible for initialising them, rather than n-myft-ui assuming they're needed and setting up the event listener regardless.

dan-searle:
* Make pin button init safe to call multiple times
* Pin button init will init `next-myft-client` itself (which is safe to do multiple times)
* myft-buttons init will no longer init the pin-buttons - consumers will need to do that explicitly

Fixes #241 

**Breaking change!** Consumers of `n-myft-ui` that need pin buttons will need to call `n-myft-ui/components/pin-button/index.js` default export explicitly.

### DIFF
--- a/components/pin-button/index.js
+++ b/components/pin-button/index.js
@@ -33,13 +33,17 @@ const togglePrioritised = (conceptId, prioritised) => {
 };
 
 export default () => {
-	delegate.on('click', 'button[data-prioritise-button]', event => {
-		event.preventDefault();
+	myftApiClient.init()
+		.then(() => {
+			delegate.off();
+			delegate.on('click', 'button[data-prioritise-button]', event => {
+				event.preventDefault();
 
-		const { conceptId, prioritised } = event.target.dataset;
-		const wrapper = findAncestor(event.target, 'myft-pin-button-wrapper');
+				const { conceptId, prioritised } = event.target.dataset;
+				const wrapper = findAncestor(event.target, 'myft-pin-button-wrapper');
 
-		setLoading(wrapper);
-		togglePrioritised(conceptId, prioritised === 'true');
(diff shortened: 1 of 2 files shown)


components/pin-button/index.js
```diff
@@ -33,13 +33,17 @@ const togglePrioritised = (conceptId, prioritised) => {
...
-
-		const { conceptId, prioritised } = event.target.dataset;
-		const wrapper = findAncestor(event.target, 'myft-pin-button-wrapper');
-
-		setLoading(wrapper);
-		togglePrioritised(conceptId, prioritised === 'true');
-	});
+	myftApiClient.init()
+		.then(() => {
+			delegate.off();
```
benbarnett: Would this unset any other event listeners with this delegate?

dan-searle: Yes, but this delegate instance is created in this file and isn't used for anything else, so it's safe.

benbarnett: Ohh yes so it is. Cool. Does `myftApiClient.init()` resolve if called multiple times? Just thinking that this takes a slightly different strategy to the myft-buttons init (which uses an `initialised` variable to handle this case)

PR #244, opened 2018-09-28
Modify unread articles indicator - chronology for cross-device

fenglish:
🐿 v2.10.3

determinenewarticlesincetime2

In `chronology.spec.js`, we use [timekeeper](https://www.npmjs.com/package/timekeeper) instead of `sinon.useFakeTimers`.

We have found something wrong on `sinon.useFakeTimers` on IE11, however we don't know what exactly. All the tests passed after the change.

**The fail test's job** => https://circleci.com/gh/Financial-Times/n-myft-ui/1835


components/unread-articles-indicator/chronology.js
```diff
@@ -1,30 +1,29 @@
-import { differenceInMinutes, isAfter, startOfDay } from 'date-fns';
+import { differenceInMinutes, isAfter, isToday, startOfDay } from 'date-fns';
+import fetchUserLastVisitedAt from './fetch-user-last-visited-timestamp';
```
dan-searle: Has this file been missed off the commit?

components/unread-articles-indicator/chronology.js
```diff
@@ -1,30 +1,29 @@
...
+	if (isToday(userNewArticlesSince) && dateIsWithinSameVisitThreshold(userNewArticlesSince)) {
+		return userNewArticlesSince;
+	}
+
 	const earliestNewArticlesSince = startOfDay(new Date()).toISOString();
 
-	if (!isValidPublishedSince(userLastVisitedAt, earliestNewArticlesSince)) {
-		return earliestNewArticlesSince;
-	}
+	return fetchUserLastVisitedAt(uuid)
```
dan-searle: Because `determineNewArticlesSinceTime` is no longer synchronous, it'll be a breaking change.

[index.js](https://github.com/Financial-Times/n-myft-ui/blob/master/components/unread-articles-indicator/index.js#L39) will need to be changed to only update localStorage when this determineNewArticlesSinceTime promise resolves (and to pass in the user UUID).

components/unread-articles-indicator/chronology.js
```diff
@@ -1,30 +1,29 @@
...
-
-const isValidPublishedSince = (dateToValidate, defaultPublishedSince) => {
-	return typeof dateToValidate === 'string' && isAfter(dateToValidate, defaultPublishedSince);
-};
-
 const dateIsWithinSameVisitThreshold = date => differenceInMinutes(new Date(), date) <= SAME_VISIT_THRESHOLD_MINUTES;
 
 /**
  * @param {string} userLastVisitedAt    ISO date representing when a user last visited ft.com
  * @param {string} userNewArticlesSince  ISO date representing the time we last used to determine if articles are new for the user
```
dan-searle: This param jsdoc should be updated to explain `uuid`

components/unread-articles-indicator/chronology.js
fenglish: @dan-searle I haven't done the part yet

dan-searle: Ah!

components/unread-articles-indicator/chronology.js
fenglish: @dan-searle Do you think this is ok? I haven't sorted out their tests, I will do that if this looks ok.
https://github.com/Financial-Times/n-myft-ui/pull/244/commits/55c655bba3c90b05ba1352336080a9e38f67636a

components/unread-articles-indicator/chronology.js
dan-searle: It might be worth combining the new file with `fetch-new-article.js` (and renaming it something more general - `api.js`?) - to contain all code that interacts with services.

components/unread-articles-indicator/index.js
```diff
@@ -34,14 +34,16 @@ const showUnreadArticlesCount = ({ uuid, newArticlesSinceTime, withTracking = fa
...
-		storage.setLastVisitedAt();
+		return determineNewArticlesSinceTime(storage.getNewArticlesSinceTime(), uuid)
+			.then(timestamp => {
+				storage.setNewArticlesSinceTime(timestamp);
+				return { uuid, newArticlesSinceTime: timestamp };
+			});
 	}
 
-	return newArticlesSinceTime;
+	return { uuid, newArticlesSinceTime };
```
dan-searle: I think this function should only return `newArticlesSinceTime` (not `uuid` as well).

components/unread-articles-indicator/index.js
```diff
@@ -55,9 +57,10 @@ export default () => {
 	});
 
 	return getUserId
-		.then(uuid => showUnreadArticlesCount({
+		.then(uuid => getNewArticlesSinceTime(uuid))
+		.then(({ uuid, newArticlesSinceTime }) => showUnreadArticlesCount({
```
dan-searle: After `getNewArticlesSinceTime` is changed to only return the `newArticleSinceTime`, this will need to be handled differently.

When you need a value from early in a promise chain, later on in the chain, you can do that by breaking the chain into parts and using Promise.all to put it back together - something like this:

```es6
const userIdPromise = sessionClient.uuid().then(({ uuid }) => uuid);
const newArticleSincePromise = getUserId.then(uuid => getNewArticlesSinceTime);

return Promise.all([userIdPromise, newArticleSincePromise])
	.then(([uuid, newArticleSinceTime]) => showUnreadArticlesCount({
		uuid,
		newArticleSinceTime,
		withTracking: true
	}))
```

components/unread-articles-indicator/index.js
dan-searle: See this comment: https://github.com/Financial-Times/n-myft-ui/pull/244#pullrequestreview-164274643

components/unread-articles-indicator/chronology.js
dan-searle: See this comment: https://github.com/Financial-Times/n-myft-ui/pull/244#pullrequestreview-164274643

components/unread-articles-indicator/index.js
fenglish: 👍

components/unread-articles-indicator/chronology.js
fenglish: 👍

components/unread-articles-indicator/chronology.js
fenglish: @dan-searle @benbarnett Is it ok if I make change to contain all code that interacts with api into `api.js` after this PR? If I include the change in this PR, it would be a long PR because I need to change their tests as well.

components/unread-articles-indicator/index.js
```diff
@@ -54,19 +56,23 @@ export default () => {
 		}
 	});
 
-	return getUserId
-		.then(uuid => showUnreadArticlesCount({
+	const userIdPromise = getUserId;
```
dan-searle: You could delete line 50 and change this to `const userIdPromise = sessionClient.uuid().then(({ uuid }) => uuid);`

components/unread-articles-indicator/index.js
```diff
@@ -54,19 +56,23 @@ export default () => {
...
+
+	return Promise.all([userIdPromise, newArticleSincePromise])
+		.then(([uuid, newArticlesSinceTime]) => showUnreadArticlesCount({
 			uuid,
-			newArticlesSinceTime: getNewArticlesSinceTime(),
+			newArticlesSinceTime,
 			withTracking: true
 		}))
 		.then(() => {
 			document.addEventListener('visibilitychange', () => {
```
fenglish: I'm not sure what this part is for. It runs showUnreadArticlesCount when something visible...?

dan-searle: It's when you've left the page in a background (or non-visible tab/window), and then switch back to it.

fenglish: I see!!

components/unread-articles-indicator/index.js
fenglish: 👍

components/unread-articles-indicator/index.js
```diff
@@ -34,18 +34,20 @@ const showUnreadArticlesCount = ({ uuid, newArticlesSinceTime, withTracking = fa
...
-		storage.setLastVisitedAt();
+		return determineNewArticlesSinceTime(storage.getNewArticlesSinceTime(), uuid)
+			.then(timestamp => {
+				storage.setNewArticlesSinceTime(timestamp);
+				newArticlesSinceTime = timestamp;
+				return newArticlesSinceTime;
+			});
 	}
 
 	return newArticlesSinceTime;
```
benbarnett: Can we return this with `Promise.resolve()` so this function always return a promise?

test/unread-articles-indicator/chronology.spec.js
```diff
@@ -1,100 +1,126 @@
...
 			beforeEach(() => {
 				userLastVisitedAt = TODAY_0800;
-				userNewArticlesSince = TODAY_0700;
-				timeNow = new Date(TODAY_0801);
-				clock = sinon.useFakeTimers(timeNow);
+				userNewArticlesSince = TODAY_0750;
 			});
 
-			describe('and there is a valid userNewArticlesSince time set', () => {
+			describe('and returns within the same-visit thresholdand', () => {
```
benbarnett: slight typo here

components/unread-articles-indicator/index.js
fenglish: 👍

test/unread-articles-indicator/chronology.spec.js
fenglish: 🙈

components/unread-articles-indicator/device-session.js
```diff
@@ -0,0 +1,23 @@
...
+
+class DeviceSession {
+
+	constructor () {
+		this.expiry = storage.getDeviceSessionExpiry();
+		const newExpiry = addMinutes(new Date(), SESSION_THRESHOLD_MINUTES).toISOString();
+		storage.setDeviceSessionExpiry(newExpiry);
+	}
+
+	isWithinSession (targetTimestamp) {
```
dan-searle: It's a minor, pedantic and opinionated point, but I think re-framing in the inverse (something like `isNewSession`) makes more sense to me - as a consumer of this we don't actually care about the session itself, just whether a new one has started.

Also, more importantly, I don't think we should have to pass anything into this function, as it can just check against `new Date()`?

components/unread-articles-indicator/device-session.js
```diff
@@ -0,0 +1,23 @@
...
+		if (this.expiry) {
+			return isBefore(targetTimestamp, this.expiry) ? true : false;
```
dan-searle: The ternary is redundant - this does the same thing:

```suggestion
			return isBefore(targetTimestamp, this.expiry);
```

(`isBefore` returns a boolean anyway)

test/unread-articles-indicator/chronology.spec.js
```diff
@@ -1,100 +1,132 @@
...
 			beforeEach(() => {
 				userLastVisitedAt = TODAY_0800;
-				userNewArticlesSince = TODAY_0700;
-				timeNow = new Date(TODAY_0801);
-				clock = sinon.useFakeTimers(timeNow);
+				userNewArticlesSince = TODAY_0750;
 			});
 
-			describe('and there is a valid userNewArticlesSince time set', () => {
+			describe('and returns within DeviceSessionExpiry', () => {
```
dan-searle: Might be better to avoid referring to `DeviceSession` in the description here, and be more agnostic like "and returns within the same session"

test/unread-articles-indicator/chronology.spec.js
```diff
@@ -1,100 +1,132 @@
...
+						.then(newArticlesSinceTime => {
+							expect(newArticlesSinceTime).to.equal(userNewArticlesSince);
+						});
 				});
 			});
 
-			describe('and there is no (or an invalid) userNewArticlesSince time set', () => {
-				it('should return the EARLIEST_NEW_ARTICLES_TIME', () => {
-					const newArticlesSinceTime = determineNewArticlesSinceTime(userLastVisitedAt, null);
+			describe('and returns after DeviceSessionExpiry', () => {
```
dan-searle: ...and here

components/unread-articles-indicator/device-session.js
fenglish: Yeah. Sorry... passing `newArticleSince` doesn't make sense... My brain is exploding :dizzy_face:

dan-searle: No worries, this chronology stuff has that effect!

test/unread-articles-indicator/device-session.spec.js
```diff
@@ -0,0 +1,88 @@
...
+				expect(deviceSession.isNewSession()).to.be.true;
+			});
+
+			it('and targetTimestamp is after this.expiry, should return false', () => {
+				clock = sinon.useFakeTimers(new Date(timestampAfterExpiry));
+				expect(deviceSession.isNewSession()).to.be.false;
+			});
+		});
+
+		context('if this.expiry is undefined', () => {
```
dan-searle: In this scenario, wouldn't we expect `isNewSession` to return `true`?

test/unread-articles-indicator/device-session.spec.js
```diff
@@ -0,0 +1,88 @@
...
+
+	describe('isNewSession:', () => {
+
+		context('if this.expiry is defined', () => {
+			beforeEach(() => {
+				deviceSessionExpiry = expiryTimestamp;
+				deviceSession = new DeviceSession();
+			});
+
+			it('and targetTimestamp is before this.expiry, should return true', () => {
```
dan-searle: I'm not sure what `targetTimestamp` is in this and the following test?

test/unread-articles-indicator/device-session.spec.js
```diff
@@ -0,0 +1,88 @@
...
+			const correctTimestamp = dateFns.addMinutes(timeNow, SESSION_THRESHOLD_MINUTES).toISOString();
+
+			expect(mockStorage.setDeviceSessionExpiry).to.have.been.calledOnce;
+			expect(mockStorage.setDeviceSessionExpiry).to.have.been.calledWith(correctTimestamp);
+		});
+	});
+
+	describe('isNewSession:', () => {
+
+		context('if this.expiry is defined', () => {
```
dan-searle: These tests are very specific to the current implementation of this class and I think that makes them a bit confusing. They would be better based around the use cases the code needs to support:

* **given a user visits for the first time** (initially no value in localStorage)
  * `this.isNewSession()` should return `true`
  * should set the localStorage value to now + 30 minutes
* **given a user returns after 30 minutes** (initial value in localStorage is in the past)
  * `this.isNewSession()` should return `true`
  * should set the localStorage value to now + 30 minutes
* **given a user returns within 30 minutes** (initial value in localStorage is in the future)
  * `this.isNewSession()` should return `false`
  * should set the localStorage value to now + 30 minutes

I don't think you need to assert anything about `this.expiry`, as that's an implementation detail rather than part of this class's API.

components/unread-articles-indicator/device-session.js
```diff
@@ -0,0 +1,23 @@
...
+
+	constructor () {
+		this.expiry = storage.getDeviceSessionExpiry();
+		const newExpiry = addMinutes(new Date(), SESSION_THRESHOLD_MINUTES).toISOString();
+		storage.setDeviceSessionExpiry(newExpiry);
+	}
+
+	isNewSession () {
+		if (this.expiry) {
+			return isBefore(new Date(), this.expiry);
```
dan-searle: I think this should be `isAfter` now?

fenglish: Oh.... 🙈 why tests passed. the tests are wrong as well, aren't they 😭

test/unread-articles-indicator/device-session.spec.js
fenglish: :flushed:

test/unread-articles-indicator/device-session.spec.js
fenglish: 👍

components/unread-articles-indicator/chronology.js
```diff
@@ -1,16 +1,16 @@
...
+import DeviceSession from './device-session';
 
 /**
  * @param {string} userNewArticlesSince  ISO date representing the time we last used to determine if articles are new for the user
  * @param {string} uuid  User uuid
  * @return {string} ISO date when we now determine articles to be 'new' for the user
  */
 export const determineNewArticlesSinceTime = (userNewArticlesSince, uuid) => {
-	if (isToday(userNewArticlesSince) && dateIsWithinSameVisitThreshold(userNewArticlesSince)) {
+	const deviceSession = new DeviceSession();
```
benbarnett: I think we should avoid creating a new instance of this every time. Probably once per page is enough?

PR #248, opened 2018-10-16
Fix word-break in concept-list

kateloschinina:
## What

Based on the issue
https://trello.com/c/riNfp87l/400-break-word-issue-on-side-concept-list

### Before
screen shot 2018-10-16 at 14 26 18

### After
screen shot 2018-10-16 at 14 26 56

### Browser support
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

### DIFF
--- a/components/concept-list/main.scss
+++ b/components/concept-list/main.scss
@@ -34,7 +34,7 @@
 }
 
 .concept-list__concept {
-	word-break: break-all;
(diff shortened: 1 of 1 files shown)


kateloschinina:
[Break word issue on side concept-list](https://trello.com/c/riNfp87l/400-break-word-issue-on-side-concept-list)

PR #249, opened 2018-10-19
do not display indicator if localStorage is not available

fenglish:
🐿 v2.10.3

fix these errors
https://sentry.io/nextftcom/ft-next-article/issues/617368177/?query=is:unresolved
https://sentry.io/nextftcom/ft-next-article/issues/617369535/?query=is:unresolved

### DIFF
--- a/components/unread-articles-indicator/index.js
+++ b/components/unread-articles-indicator/index.js
@@ -45,6 +45,10 @@ export const getNewArticlesSinceTime = () => {
 };
 
 export default () => {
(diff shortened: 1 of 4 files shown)


dan-searle:
We'll need to make a change to `next-myft-page`'s [latest-view component](https://github.com/Financial-Times/next-myft-page/blob/master/client/components/latest-view/data.js) to use the new `storage.isAvailable()` function too.

PR #253, opened 2018-11-02
Passes the form data through on enhanced pinning

GlynnPhillips:
The data is put into the form but for enhanced journeys the form is
hijacked and the csrf token wasn't being passed through

### DIFF
--- a/components/pin-button/index.js
+++ b/components/pin-button/index.js
@@ -40,7 +40,7 @@ export default () => {
 			delegate.on('click', 'button[data-prioritise-button]', event => {
 				event.preventDefault();
 
-				const form = findAncestor(event.target, 'myftPrioritise');
+				const form = event.target.closest('[data-myft-prioritise]');
 				const formData = form ? serialize(form, { hash: true }) : {};
 				const { conceptId, prioritised } = event.target.dataset;
 				const wrapper = findAncestor(event.target, 'myft-pin-button-wrapper');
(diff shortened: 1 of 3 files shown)


components/pin-button/index.js
```diff
@@ -13,17 +14,17 @@ const trackPinningAction = ({ action }) =>
 		bubbles: true
 	});
 
-const findAncestor = (el, classname) => el.classList.contains(classname)
+const findAncestor = (el, selector) => el.classList.contains(selector) || el.dataset[selector] !== undefined
```
chee: this feels a little eek, maybe this should be two different functions.

actually, we use `default` from polyfill.io which includes `Element.prototype.closest` so we can use that instead!

components/pin-button/index.js
```diff
@@ -39,11 +40,13 @@ export default () => {
 			delegate.on('click', 'button[data-prioritise-button]', event => {
 				event.preventDefault();
 
+				const form = findAncestor(event.target, 'myftPrioritise');
```
chee:
```suggestion
				const form = event.target.closest('[data-myft-prioritise]');
```

components/pin-button/index.js
```diff
@@ -39,11 +40,13 @@ export default () => {
...
+				const formData = form ? serialize(form, { hash: true }) : {};
 				const { conceptId, prioritised } = event.target.dataset;
 				const wrapper = findAncestor(event.target, 'myft-pin-button-wrapper');
```
chee:
```suggestion
				const wrapper = event.target.closest('.myft-pin-button-wrapper');
```

components/pin-button/index.js
GlynnPhillips: Good shout

PR #271, opened 2018-12-18
Fixes syntax error on Object.assign

GlynnPhillips:
This meant that not all the data needed was sent in the request and meant that users were not being assigned to an email group

### DIFF
--- a/components/digest-promo/index.js
+++ b/components/digest-promo/index.js
@@ -75,7 +75,7 @@ function addToDigest () {
 		}
 	};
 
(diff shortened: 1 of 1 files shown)


components/digest-promo/index.js
```diff
@@ -75,7 +75,7 @@ function addToDigest () {
 		}
 	};
 
-	const promises = [myFtClient.add('user', null, 'preferred', 'preference', 'email-digest', Object.assign({}, {token: csrfToken}), metaEmail)];
+	const promises = [myFtClient.add('user', null, 'preferred', 'preference', 'email-digest', Object.assign({}, {token: csrfToken}, metaEmail)];
```
kateloschinina: ooops, there is one more `)` missing apparently...

PR #288, opened 2019-03-29
Offer Instant Alerts with push notifications

alexandercurtis:
Part of https://trello.com/c/duzjIVjX/3259-browser-notification-on-boarding

This copies the service worker code from `next-myft-page` into `n-myft-ui` so it can be shared across applications. 

It also adds a new UI component, offering the user Instant Alerts, and on the back of that, prompting them to enable Push Notifications.

image

Enable by setting `myftOfferInstantAlertNotifications` in the flags object of the options passed to `n-myft-ui/init(...)` (This normally means just set the `myftOfferInstantAlertNotifications` flag.)

 🐿 v2.12.0

### DIFF
--- a/myft/ui/lib/push-notifications.js
+++ b/myft/ui/lib/push-notifications.js
@@ -1,0 +1,179 @@
+import instantAlertsConfirmation from '../../../components/instant-alerts-confirmation';
+import myftClient from 'next-myft-client';
+import getToken from '../lib/get-csrf-token';
+
+const csrfToken = getToken();
+const isLocalOrHTTPS = document.location.protocol === 'https:' ||
+	document.location.href.indexOf('local') >= 0;
+
+let isServiceWorkerInitialised = false;
+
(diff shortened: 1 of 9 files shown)


myft/ui/lib/push-notifications.js
```diff
@@ -0,0 +1,179 @@
+import instantAlertsConfirmation from '../../../components/instant-alerts-confirmation';
+import myftClient from 'next-myft-client';
+import getToken from '../lib/get-csrf-token';
+
+const csrfToken = getToken();
+const isLocalOrHTTPS = document.location.protocol === 'https:' ||
+	document.location.href.indexOf('local') >= 0;
+
+let serviceWorkerInitialisedOk = false;
```
binaryberry: I _think_ it's a convention to give variable names that are going to have boolean values a name with a verb? Like `serviceWorkerIsInitialised` or `serviceWorkerHasInitialised`?

alexandercurtis: good point :+1:

PR #292, opened 2019-04-08
Add fdtomdelegate resolution

alexandercurtis:
ftdomdelegate version has been bumped, which has confused bower

More here: https://financialtimes.slack.com/archives/C041V9QA7/p1554392891109300

 🐿 v2.12.0

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -26,5 +26,8 @@
     "fetchres": "^1.7.2",
     "o-tooltip": "^3.1.2",
     "js-cookie": "^2.2.0"
+  },
+  "resolutions": {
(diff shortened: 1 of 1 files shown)


notlee:
Just for context in case you're not already aware: The conflict was caused because a new major of `n-ui-foundations` was released with the latest version of `fdtomdelegate` before Origami components where updated -- sorry if that caused you problems! The new major of `fdtomdelegate` was a breaking release due to a name change, but the api didn't change so a resolution is ok. 👍 

I'm working  updating Origami components to remove this conflict, when it's done this resolution can be removed. 😊

PR #293, opened 2019-04-15
Track appearances of instant alerts message box

alexandercurtis:
Adds tracking event `offer-instant-alerts` that fires when the message box is shown. It has a field called `permission` which reflects whether or not the user has push notifications blocked.

https://trello.com/c/lV8PhYrz/3490-track-push-notifications

 🐿 v2.12.0

### DIFF
--- a/myft/ui/lib/push-notifications.js
+++ b/myft/ui/lib/push-notifications.js
@@ -162,6 +162,16 @@ export function unsubscribe () {
 
 export function offerInstantAlerts (conceptId) {
 	if (isServiceWorkerInitialised) {
+
+		document.body.dispatchEvent(new CustomEvent('oTracking.event', {
+			detail: {
+				category: 'myFT',
+				action: 'offer-instant-alerts',
+				permission: Notification.permission
+			},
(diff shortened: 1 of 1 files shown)


myft/ui/lib/push-notifications.js
```diff
@@ -162,6 +162,16 @@ export function unsubscribe () {
 
 export function offerInstantAlerts (conceptId) {
 	if (isServiceWorkerInitialised) {
+
+		document.body.dispatchEvent(new CustomEvent('oTracking.event', {
+			detail: {
+				category: 'myFT',
+				action: 'offer-instant-alerts',
+				permission: Notification.permission
```
binaryberry: should there also be a system key in there? https://github.com/Financial-Times/o-tracking/blob/master/docs/event.md

alexandercurtis: Hmm, I can't find `system` used anywhere else. I think the documentation is referring to the event that gets sent to Spoor, rather than the DOM event.

https://github.com/Financial-Times/o-tracking#events

PR #297, opened 2019-05-23
Add tooltip to pin button on hover and focus

maxcbc:
Adds tooltip to pin button on hover and focus 

Screenshot 2019-05-23 at 16 14 04

🐿 v2.12.3


binaryberry:
Approved pending passing CircleCI tests

PR #299, opened 2019-05-30
Shift focus to subscribe/sign popup notification

ker-an:
If a user is not logged in and visits a free article, a 'Please
subscribe or sign in...' message appears over the header when they
attempt to add the article to myFT.

Previously, when the message appeared the focus would remain present on
the 'Add to myFT' button and would not shift to the message. This is a
problem for keyboard users.

This PR fixes this by shifting the focus to the message so it's
accessible to keyboard users. It's done by passing in the new
n-notification `focusSelector` and `returnFocusSelector` properties.

 🐿 v2.12.3

### DIFF
--- a/myft/ui/myft-buttons/init.js
+++ b/myft/ui/myft-buttons/init.js
@@ -25,8 +25,8 @@ function anonEventListeners () {
 	const subscribeUrl = '/products?segID=400863&segmentID=190b4443-dc03-bd53-e79b-b4b6fbd04e64';
 	const signInLink = `/login${currentPath.length ? `?location=${currentPath}` : ''}`;
 	const messages = {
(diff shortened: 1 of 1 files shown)


ker-an:
Related to: https://github.com/Financial-Times/n-notification/pull/61

myft/ui/myft-buttons/init.js
```diff
@@ -35,7 +35,9 @@ function anonEventListeners () {
 
 			nNotification.show({
 				content: messages[action],
-				trackable: 'myft-anon'
+				trackable: 'myft-anon',
+				focusSelector: '.myft-ui-subscribe',
```
sjparkinson: Will there only ever be one element with this class to select? If so, could you use an ID instead?

ker-an: Spoke to Sam in real life and we've got two elements that use this class, so we'll stick with classes.

PR #308, opened 2019-07-19
Alternative skin for collections

alexandercurtis:
This is to support the upcoming industry recommendations box.

image

https://trello.com/c/8P9xkdxf/3625-topic-discovery-personalized-industry-collections-on-myft-topic-view

### DIFF
--- a/components/collections/collections.html
+++ b/components/collections/collections.html
@@ -1,18 +1,25 @@
 <section
-	class="collection"
+	class="collection {{#if liteStyle}}collection--lite{{else}}collection--regular{{/if}}"
 	data-trackable="collection">
-	<header class="collection__header">
-		<h2 class="collection__title">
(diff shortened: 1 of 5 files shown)


components/collections/collections.html
```diff
@@ -1,18 +1,25 @@
...
+				{{#if ../liteStyle}}
+					{{>	n-myft-ui/components/follow-button/follow-button
+						variant="primary"
+						buttonText=name
+					}}
+				{{else}}
+					{{>	n-myft-ui/components/follow-button/follow-button
+						variant="inverse"
+						buttonText=name
+					}}
```
alexandercurtis: If anyone can tell me how to do change just the variant property in handlebars without repeating the whole partial, please let me know!

maxcbc: I think you can do the if statement in the property value

alexandercurtis: @maxcbc I tried but it didn't work. But I might have got the syntax wrong.

PR #316, opened 2019-08-21
Fix to DAC_Status_message_only_temporary_Issue1

thurstontye:
Duration of notification now set to 0 so user is required to dismiss the notification as per the DAC recommendation.


thurstontye:
Added documentation around the use of '0' as a value in n-notifications.

PR #317, opened 2019-08-30
Add myft header logo tooltip

maxcbc:
# Add myft header logo tooltip

## What
Add myft header logo tooltip if:
- user has myFT_HeaderTooltip flag enabled
- user has navigated from a non FT domain
- user is not on a `/myft*` page
 🐿 v2.12.3
## Mobile
Screenshot 2019-08-30 at 09 41 25

## Desktop
Screenshot 2019-08-30 at 09 41 40

### DIFF
--- a/components/header-tooltip/main.scss
+++ b/components/header-tooltip/main.scss
@@ -1,0 +1,12 @@
+.o-header__top-link--myft__container {
+	position: relative;
+
+	.o-tooltip {
+		white-space: normal;
+		text-align: left;
+		min-width: 200px;
+		@include oGridRespondTo('L') {
+			min-width: 250px;
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


components/header-tooltip/main.scss
```diff
@@ -0,0 +1,12 @@
+.o-header__top-link--myft__container {
+	position: relative;
+
+	.o-tooltip {
+		white-space: normal;
+		text-align: left;
```
thurstontye: Probably not a biggy but maybe check with origami if this alignment adhears to their conventions.

maxcbc: I think this is because by default it doesn't wrap, now i've made it do so it's more obvious.

PR #326, opened 2019-09-17
Add two variants for wording of Add to myFT button

binaryberry:
This adds two variants for the wording of the Add to myFT button, behind the `followmyFTABTest` flag.

**Control** 

Screenshot 2019-09-17 at 15 51 57

**Variant 1**

Screenshot 2019-09-17 at 15 51 37

**Variant 2**

Screenshot 2019-09-17 at 15 50 01

**Note:** I did not update the pop up as it seemed outside the scope of the card but the wording will be a bit inconsistent with the variants.

Screenshot 2019-09-17 at 15 56 55

https://trello.com/c/Z8MFkTI1/3753-myft-debranding-prepare-a-b-test

### DIFF
--- a/components/collections/collections.html
+++ b/components/collections/collections.html
@@ -63,22 +63,38 @@
 				{{~/concepts~}}"
 			/>
 			<button
-				type="submit"
-				aria-label="Add all topics in the {{title}} collection to my F T"
-				aria-pressed="false"
-				class="collection-follow-all__button {{#if liteStyle}}collection-follow-all__button--lite{{else}}collection-follow-all__button--regular{{/if}}"
-				data-alternate-label="Remove all topics in the {{title}} collection from my F T"
-				data-alternate-text="Added"
-				data-trackable="follow all"
-				data-concept-id="
-					{{~#concepts~}}
-						{{conceptId}}
-						{{~#unless @last~}}
-							,
-						{{~/unless~}}
-					{{~/concepts~}}"
-				title="Add all topics in the {{title}} collection to my F T">
-				Add all to myFT
+					type="submit"
+					aria-pressed="false"
+					class="collection-follow-all__button {{#if liteStyle}}collection-follow-all__button--lite{{else}}collection-follow-all__button--regular{{/if}}"
+					data-trackable="follow all"
+					data-concept-id="
+						{{~#concepts~}}
+							{{conceptId}}
+							{{~#unless @last~}}
+								,
+							{{~/unless~}}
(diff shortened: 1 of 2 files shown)


components/collections/collections.html
```diff
@@ -63,22 +63,38 @@ <h2 class="collection__title {{#if liteStyle}}collection__title--lite{{else}}col
...
+					Add all to feed
+				{{/ifEquals}}
+				{{#ifEquals @root.flags.followmyFTABTest 'Variant2' }}
+					aria-label="Follow all topics in the {{title}} collection"
+					data-alternate-label="Unfollow all topics in the {{title}} collection"
+					data-alternate-text="Following"
+					title="Follow all topics in the {{title}} collection">
+					Follow all
+				{{/ifEquals}}
+				{{#ifEquals @root.flags.followmyFTABTest false }}
```
alexandercurtis: I think the flag could also be `undefined`. Rather than go off to look whether that can ever happen, I would just put all this in an `if ... else if ... else` so that something gets output no matter what the value. Just makes it a bit more robust in the face on the unexpected.

components/follow-button/follow-button.html
```diff
@@ -34,33 +34,95 @@
...
+						data-alternate-text="{{alternateText}}"
+					{{else}}
+						{{#if buttonText}}
+							data-alternate-text="{{buttonText}}"
+						{{else}}
+							data-alternate-text="Unfollowed"
+						{{/if}}
+					{{/if}}
+				{{/ifEquals}}
+				{{#ifEquals @root.flags.followmyFTABTest false}}
```
alexandercurtis: see previous comment about `if else if else`

components/follow-button/follow-button.html
```diff
@@ -74,11 +136,27 @@
...
+					{{~/ifAll~}}
+				{{~/ifEquals~}}
+				{{~#ifEquals @root.flags.followmyFTABTest 'Variant2'~}}
+					{{~#ifAll setFollowButtonStateToSelected @root.cacheablePersonalisedUrl~}}
+						Following
+					{{~else~}}
+						Follow
+					{{~/ifAll~}}
+				{{~/ifEquals~}}
+				{{~#ifEquals @root.flags.followmyFTABTest false ~}}
```
alexandercurtis: ditto `if else if else`

components/collections/collections.html
binaryberry: Thanks! If/else revamped, and tests passing!

PR #391, opened 2019-11-22
remove npm dependencies in favour of bower dependencies

maxcbc:
## What
- removes npm dependencies
- adds bower dependencies
- updates `date-fns` references to resolve to bower `date-fns` dependency
- adds pre-test script to prevent addition of npm dependencies
- removes unused lodash dependency

## Why
This library is consumed as a bower dependency (so that its own bower dependencies are downloaded along side it). As a result it's npm dependencies won't be resolved properly by n-ui. This has caused deployment problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "homepage": "https://github.com/Financial-Times/n-myft-ui#readme",
   "devDependencies": {
-    "@financial-times/n-gage": "^3.11.2",
+    "@financial-times/n-gage": "^3.12.0",
     "@financial-times/n-heroku-tools": "8.3.1",
     "@financial-times/n-internal-tool": "2.3.4",
     "@financial-times/n-webpack": "^3.0.2",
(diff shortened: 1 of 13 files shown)


scripts/prevent-npm-deps.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+const {dependencies = {}} = require('../package');
+
+if (Object.keys(dependencies).length > 0) {
+	global.console.error('NPM dependencies included in bower only js project, please remove');
```
robsquires: nice 👌

package.json
```diff
@@ -77,19 +78,12 @@
...
     "sinon-chai": "^3.2.0",
-    "snyk": "^1.216.5",
-    "babel-preset-env": "^1.7.0",
-    "regenerator-runtime": "^0.13.3"
-  },
-  "dependencies": {
-    "date-fns": "2.7.0",
-    "form-serialize": "^0.7.2",
-    "js-cookie": "^2.2.0",
-    "lodash": "^4.17.10"
```
robsquires: guess this wasn't being used anyway?

maxcbc: well they were but inconsistently, if another dependency of a consuming project had these as dependencies then yes. Module resolution fun.

maxcbc: They're all (apart from lodash) now included as bower dependencies.

robsquires: Ideal

thurstontye:
@maxcbc can you hold off releasing this. It makes sense to simpliy but I'm not entirely sure this is the correct stragegy. Have you discussed this with @i-like-robots and core ui? I want to be sure this aligned with the chnaged they are making around pagekit and n-ui before releasing.

maxcbc:
@thurstontye I've already done a beta release but will hold off rolling out further

PR #396, opened 2019-11-27
rename unread articles polling flag 🐿 v2.12.5

maxcbc:
This PR renames the flag which activates the polling feature on the unread articles indicator.

This is being done so that outdated code (which caused traffic issues) is not activated when we turn on the feature


robsquires:
As discussed IRL, a bit of context around this will be useful if we need to look back at this

PR #397, opened 2019-11-29
Unread articles refinement

maxcbc:
- update dismiss logic so that it updates the last time you visited the myft feed page every time you go to the myft feed page, not just when you click the myft button

### DIFF
--- a/components/unread-articles-indicator/index.js
+++ b/components/unread-articles-indicator/index.js
@@ -11,7 +11,7 @@ let updateTimeout;
 let initialFeedStartTime;
 let userId;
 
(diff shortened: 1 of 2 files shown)


components/unread-articles-indicator/index.js
```diff
@@ -11,14 +11,15 @@ let updateTimeout;
 let initialFeedStartTime;
 let userId;
 
+const isMyftFeedPage = window.location.pathname.startsWith('/myft/following');
```
robsquires: As discussed IRL, don't think IE supports this

PR #403, opened 2019-12-09
bump update interval to 10mins 🐿 v2.12.5

thurstontye:
Increase UPDATE_INTERVAL from 5mins to 10mins to reduce the rate for requests we generate to the session-service-api.

### DIFF
--- a/components/unread-articles-indicator/README.md
+++ b/components/unread-articles-indicator/README.md
@@ -23,9 +23,9 @@ It indicates to the user that articles they've not yet read have been published 
 
 ## When?
 
-- June 2018 - intial implementation
+- June 2018 - intial implementation [AB Test results](https://sites.google.com/ft.com/ftproductanalytics/analysis/myft/myft-feed-page)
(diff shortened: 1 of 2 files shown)


alexandercurtis:
I would be inclined to go for even longer than 10 minutes.

10 mins is double the current value, so we would expect traffic to halve, from 1500 rps to 750 rps. Do we know that 750rps is acceptable?

I don't think the user experience would be badly affected if it only updated every 20 minutes, or maybe even longer than that.

robsquires:
We can observe a reduction in the requests coming from the indicator here:

```
index="restricted_ftdotcom" source="6hf3abuu92eqmlup9bfd7q"
| rex field="request.path" "source=(?<system_source>[^&]+)"
| rex field="request.path" "^(?<route>/v2/.+/)"
| search "system_source"="myft-ui"
| timechart count by system_source
```

thurstontye:
Looks like this is now stale and not relelvant

PR #404, opened 2019-12-11
Fix core experience

alexandercurtis:
Puts the non-js `action` attribute back (1st commit), and simplifies the logic (2nd commit), since `csrfToken` has no effect on what action is selected.

Looks like the non-js `action` behaviour got added here https://github.com/Financial-Times/n-myft-ui/commit/1bc2e27
and removed here https://github.com/Financial-Times/n-myft-ui/pull/206

Trello: 
https://trello.com/c/swsWxJK2/3968-myft-follow-unfollow-button-core-experience-404s


robsquires:
Seems to have simplified things a bit! 👍

PR #409, opened 2020-01-08
Fix unread articles indicator displays incorrect number on cross device

fenglish:
This PR includes
- **Fix `isNewSession` function never returns `true`**
When isNewSession never returns true, it always get the user's last visit timestamp from window.localStorage and never fetches it from VoltDB.

- **Fix `startTime` in `updateCount` function is set incorrect time**
The startTime to count unread articles should have been set user's last visit time or indicator dismissed time, whichever comes later.
However it's always set indicator dismissed time.

-----

### The cause of the two issues above is the same
When it gets dates from storage, it returns dates (`Tue Aug 19 1975 23:15:30 GMT+0100`) which parsed from ISO string(`1975-08-19T22:15:30.000Z`), however, these functions try to parse dates again as an ISO string.

```js
parseISO("Tue Aug 19 1975 23:15:30 GMT+0100");
// returns Invalid Date
```
`isAfter` function doesn't throw an error when it's passed `Invalid Date`, it just returns `false`.

These bugs maybe started after [this commit](https://github.com/Financial-Times/n-myft-ui/commit/68692da6bd9bb4f1966404fd97c885cd1a3dc685)

🐿 v2.12.5

### DIFF
--- a/components/unread-articles-indicator/date-fns.js
+++ b/components/unread-articles-indicator/date-fns.js
@@ -1,0 +1,32 @@
+// date-fns from v2 doesn't accept String arguments anymore.
+// the detail => https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md#200---2019-08-20
+// By adding validation for dates before their functions allows us to know it when unexpected value passed.
+
+import isTodayOriginal from 'date-fns/src/isToday';
+import isAfterOriginal from 'date-fns/src/isAfter';
+import addMinutesOriginal from 'date-fns/src/addMinutes';
+import startOfDayOriginal from 'date-fns/src/startOfDay';
+import isValidOriginal from 'date-fns/src/isValid';
+import parseISO from 'date-fns/src/parseISO';
+
+const isValid = (date) => {
+	if (!isValidOriginal(date)) {
+		console.error('Invalid date passed', [date]); //eslint-disable-line
+	}
+	return date;
+};
+
+const isToday = (date) => isTodayOriginal(isValid(date));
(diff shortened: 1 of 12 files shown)


robsquires:
Great detective work 👍 🔍 
Couple thoughts around tests..
- Is there a test case we could add to `unread-articles-indicator/update-count` for this change?
- Looks like this bug was caused by us passing around an invalid date without realising 🤔 Is there any code we could add to catch that earlier?

robsquires:
> Looks like this bug was caused by us passing around an invalid date without realising
Actually sorry, I think its more like we were passing an invalid date to `date-fns` functions without realising

components/unread-articles-indicator/date-fns.js
```diff
@@ -0,0 +1,41 @@
...
+import startOfDayOriginal from 'date-fns/src/startOfDay';
+
+import isValid from 'date-fns/src/isValid';
+import parseISO from 'date-fns/src/parseISO';
+
+const parseDate = (date) => {
+	let dateParsed;
+
+	if (typeof date === 'string') {
+		dateParsed = parseISO(date);
```
robsquires: Do still we use dates in both the `String` and `Date` type within this component?
I'm wondering if it would be better to use one or the other throughout

fenglish: @robsquires Yes, the both the `String` and `Date` are used.
When it get unread articles, it fetches articles. the articles published time stamp is `String`
https://github.com/Financial-Times/n-myft-ui/blob/master/components/unread-articles-indicator/count-unread-articles.js#L11

PR #422, opened 2020-08-19
Rewrite links to the alpha homepage

taraojo:
As part of our navigation alpha test we want to rewrite front page links so users can use the alpha front page as part of their normal routine. We check for the `frontPageAlpha` flag to do this.

The flag is set using the cookie from a separate PR: https://github.com/Financial-Times/next-myft-page/pull/2312

### DIFF
--- a/myft/ui/index.js
+++ b/myft/ui/index.js
@@ -3,11 +3,13 @@ import * as lists from './lists';
 import personaliseLinks from './personalise-links';
 import updateUi from './update-ui';
 import * as headerTooltip from '../../components/header-tooltip';
+import navigationAlphaTest from './navigationAlphaTest';
 
 function init (opts) {
 	myFtButtons.init(opts);
 	lists.init();
 	headerTooltip.init();
+	navigationAlphaTest(opts);
 }
 
 export {
--- a/myft/ui/lists.js
+++ b/myft/ui/lists.js
@@ -11,7 +11,7 @@ const delegate = new Delegate(document.body);
 const csrfToken = getToken();
 
 
-function openOverlay(html, { name = 'myft-ui', title = '&nbsp;', shaded = false }) {
+function openOverlay (html, { name = 'myft-ui', title = '&nbsp;', shaded = false }) {
 	// If an overlay already exists of the same name destroy it.
 	const overlays = Overlay.getOverlays();
 	const existingOverlay = overlays[name];
@@ -31,7 +31,7 @@ function openOverlay(html, { name = 'myft-ui', title = '&nbsp;', shaded = false 
 	});
 }
 
-function updateAfterAddToList(listId, contentId, wasAdded) {
+function updateAfterAddToList (listId, contentId, wasAdded) {
 
 	myFtUiButtonStates.setStateOfButton('contained', contentId, wasAdded);
 
@@ -54,7 +54,7 @@ function updateAfterAddToList(listId, contentId, wasAdded) {
 }
 
 
-function setUpSaveToExistingListListeners(overlay, contentId) {
+function setUpSaveToExistingListListeners (overlay, contentId) {
 
 	const saveToExistingListButton = overlay.content.querySelector('.js-save-to-existing-list');
 	const listSelect = overlay.content.querySelector('.js-list-select');
@@ -77,7 +77,7 @@ function setUpSaveToExistingListListeners(overlay, contentId) {
 	}
 }
 
-function setUpCreateListListeners(overlay, contentId) {
+function setUpCreateListListeners (overlay, contentId) {
 
 	const createListButton = overlay.content.querySelector('.js-create-list');
 	const nameInput = overlay.content.querySelector('.js-name');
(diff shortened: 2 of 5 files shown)


myft/ui/navigationAlphaTest.js
```diff
@@ -0,0 +1,13 @@
+import { $$ as findElements } from 'n-ui-foundations';
+
+export default function (opts) {
+	if (opts && opts.flags && opts.flags.get('frontPageAlpha')) {
```
8enSmith, line 4: Will this facility to override the links also be needed in the Beta or other front page variants? If so we should think about giving this functionality a more generic name as opposed to "frontPageAlpha" e.g. "linkOverrides"?

myft/ui/navigationAlphaTest.js
```diff
@@ -0,0 +1,13 @@
...
+		const links = findElements('a[href="/"], a[href^="/?"]');
```
8enSmith, line 5: As we have an "absoluteLinks" variable in this function, I suggest re-naming this to "relativeLinks" to be equally explicit.

myft/ui/navigationAlphaTest.js
```diff
@@ -0,0 +1,13 @@
...
+		const absoluteLinks = findElements('a[href="https://www.ft.com/"], a[href="http://www.ft.com/"]');
+
+		[...links, ...absoluteLinks].forEach(link => {
+			const url = new URL(link.href);
+			link.href = `https://ft-next-alpha-front-page-eu.herokuapp.com/next-alpha-front-page${url.search}`;
```
8enSmith, line 10: If this facility to override links is going to be used my multiple front page variants (e.g. alpha, beta etc.) then "https://ft-next-alpha-front-page-eu.herokuapp.com/next-alpha-front-page" should be configurable and not hard-coded.

myft/ui/navigationAlphaTest.js
8enSmith, line 5: As we have an "absoluteLinks" variable in this function, I suggest re-naming this to "relativeLinks" to be equally explicit.

myft/ui/navigationAlphaTest.js
8enSmith, line 4: Will this facility to override the links also be needed in the Beta or other front page variants? If so we should think about giving this functionality a more generic name as opposed to "frontPageAlpha" e.g. "linkOverrides" or "navigationLinkOverrides"?

myft/ui/index.js
```diff
@@ -3,11 +3,13 @@ import * as lists from './lists';
 import personaliseLinks from './personalise-links';
 import updateUi from './update-ui';
 import * as headerTooltip from '../../components/header-tooltip';
+import navigationAlphaTest from './navigationAlphaTest';
```
8enSmith, line 6: Consider renaming "navigationAlphaTest" to "navigationLinkOverrides" i.e. something more explicit. See subsequent comment too. However I realise naming is hard!

8enSmith (commented):
A few comments.

myft/ui/navigationAlphaTest.js
```diff
@@ -0,0 +1,13 @@
+import { $$ as findElements } from 'n-ui-foundations';
+
+export default function (opts) {
+	if (opts && opts.flags && opts.flags.get('frontPageAlpha')) {
+		const links = findElements('a[href="/"], a[href^="/?"]');
+		const absoluteLinks = findElements('a[href="https://www.ft.com/"], a[href="http://www.ft.com/"]');
```
8enSmith, line 6: "www.ft.com/" could be put into a constant and referenced instead of duplicating it, though I suppose there are only two instances of it!

myft/ui/navigationAlphaTest.js
taraojo: That would make sense in the long term, for now I think it's ok being hard-coded

myft/ui/navigationAlphaTest.js
taraojo: That makes sense to me 👍

myft/ui/navigationAlphaTest.js
taraojo: This makes sense to me, but again I don't think we need to worry about the other variants for now. I hadn't thought about the optional chaining syntax, do we need to be using babel for that?

myft/ui/index.js
taraojo: Haha I'm happy with either, I also wanted to make it clear that this was test code and not permanent

myft/ui/lists.js
```diff
@@ -11,7 +11,7 @@ const delegate = new Delegate(document.body);
 const csrfToken = getToken();
 
 
-function openOverlay(html, { name = 'myft-ui', title = '&nbsp;', shaded = false }) {
+function openOverlay (html, { name = 'myft-ui', title = '&nbsp;', shaded = false }) {
```
taraojo, line 14: Did these changes come from a linter? Maybe they can be excluded from this commit?

taraojo (commented):
This looks great, thanks @8enSmith, it looks like these changes would make sense to be one commit. Would you be happy to squash them?

myft/ui/lists.js
8enSmith: Yes, I didn't explicitly change this. The linter picked it up.

myft/ui/index.js
8enSmith: I'll leave the name "as is" because this is test code.

PR #423, opened 2020-08-27
Fix alpha flag access

8enSmith:
Refactored the code to use the flags.get() function to access a flag value.

### DIFF
--- a/test/navigationAlphaTest.spec.js
+++ b/test/navigationAlphaTest.spec.js
@@ -31,11 +31,14 @@ describe('navigationAlphaTest', () => {
 			{
 				flags:
 				{
-					frontPageAlpha: true
+					get: (_flag) => { return 'on'; }
 				}
 			} :
 			{
-				flags : {}
+				flags :
+				{
+					get: (_flag) => { return undefined; }
+				}
 			};
 	};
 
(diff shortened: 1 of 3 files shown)


taraojo (approved):
Thanks Ben 🎉

fenglish (approved):
Thank you!

test/navigationAlphaTest.spec.js
```diff
@@ -31,12 +31,14 @@ describe('navigationAlphaTest', () => {
...
 				flags:
 				{
+					// eslint-disable-next-line no-unused-vars
 					get: (_flag) => { return 'on'; }
 				}
 			} :
 			{
 				flags :
 				{
+					// eslint-disable-next-line no-unused-vars
```
taraojo, line 41: Is it worth removing the unused `_flag` instead of disabling the linter?

8enSmith: The get function takes an input parameter but I'm not using it. I'll look to see if I can work around it.

taraojo: I guess it gives clarity to how the function works, so we can leave this as is and merge 🙂

8enSmith: I've removed the redundant flag

PR #424, opened 2020-08-28
Add support for alternate way that frontPageAlpha flag is set

8enSmith:
This PR modifies the code so that the frontPageAlpha flag can either be read from a flags object or directly from a flags property.

### DIFF
--- a/myft/ui/navigationAlphaTest.js
+++ b/myft/ui/navigationAlphaTest.js
@@ -4,7 +4,7 @@ export default function (opts) {
 	if (
 		opts &&
 		opts.flags &&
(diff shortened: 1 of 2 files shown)


taraojo (approved):
This looks good thanks Ben 🎉 

For the record, we've had to do this because we're sending the flags instance to this package in [`next-stream-page`](https://github.com/Financial-Times/next-stream-page/blob/109dd462a5ae2aa2d8f93c185ce6e23e8ff3749c/client/main.js#L164) , while in [`next-myft-page`](https://github.com/Financial-Times/next-myft-page/blob/e05df1ae101e4653a37d479874621cfc13d792d4/client/main.js#L70) we send `flags.getAll()` which is all the flags as a standard JS object.

PR #426, opened 2020-09-04
Change URLs which get overriden

8enSmith:
This PR changes the URLs which are overridden to point to the Alpha front page URL.

Note that these overrides are only applied if the user has opted-in to the frontPageAlpha feature (this is triggered via navigating to /myft/alpha-front-page?switch=on)


test/navigationAlphaTest.spec.js
```diff
@@ -72,11 +72,16 @@ describe('navigationAlphaTest', () => {
 	};
 
 	it('Should override secure absolute links', () => {
-		baseOverrideTest('https://www.ft.com/content/a5676e20-5c92-47f3-a76c-11f9761121f5');
+		baseOverrideTest('https://www.ft.com/');
```
robsquires, line 75: While we're here, maybe cater for for homepage urls without `/` too? 
https://www.ft.com
http://www.ft.com
?

8enSmith: I've changed the code (and added tests) to cater for ft.com without trailing forward slash.

robsquires (approved):

PR #428, opened 2020-10-21
CON-440 - updated beta frontpage redirect url

gyss:
* renamed frontPageAlpha flag to betaFrontPageOptIn
* updated beta frontpage redirect url
* rewording from alpha to beta

Meta: 
[Jira](https://financialtimes.atlassian.net/browse/CON-439) + [Jira](https://financialtimes.atlassian.net/browse/CON-440)

### DIFF
--- a/myft/ui/navigationBetaTest.js
+++ b/myft/ui/navigationBetaTest.js
@@ -22,11 +22,11 @@ export default function (opts) {
 
 		const absoluteLinks = findElements(cssSelectors);
 
-		const alphaFrontPageUrl = `https://${ft}/page/home`;
+		const betaFrontPageUrl = `https://${ft}/page/home`;
 
 		[...relativeLinks, ...absoluteLinks].forEach((link) => {
 			const url = new URL(link.href);
-			link.href = `${alphaFrontPageUrl}${url.search}`;
+			link.href = `${betaFrontPageUrl}${url.search}`;
 		});
 	}
 }
--- a/myft/ui/index.js
+++ b/myft/ui/index.js
@@ -3,13 +3,13 @@ import * as lists from './lists';
 import personaliseLinks from './personalise-links';
 import updateUi from './update-ui';
 import * as headerTooltip from '../../components/header-tooltip';
-import navigationAlphaTest from './navigationAlphaTest';
+import navigationBetaTest from './navigationBetaTest';
 
 function init (opts) {
 	myFtButtons.init(opts);
 	lists.init();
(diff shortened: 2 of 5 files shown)


myft/ui/navigationBetaTest.js
```diff
@@ -4,7 +4,7 @@ export default function (opts) {
 	if (
 		opts &&
 		opts.flags &&
-		((opts.flags.get && opts.flags.get('frontPageAlpha')) || opts.flags.frontPageAlpha)
+		((opts.flags.get && opts.flags.get('betaFrontPageOptIn')) || opts.flags.betaFrontPageOptIn)
```
kavanagh, line 7: I think I want to change my mind about flag names.

I think this one should be `betaHomePage` and the kill switch should be `betaHomePageKillSwitch`

Based on a discussion I had with Anna and the data team yesterday, in the future I think there will need to be a third flag for the Opt-in button. In that case it should be something like `betaHomePageOptInJourney` so I'd rather avoid the Opt-in part of the name at the moment.

kavanagh (changes requested):
Looks good. Just change - flag names 😃

kavanagh (approved):

myft/ui/navigationBetaTest.js
```diff
@@ -4,7 +4,7 @@ export default function (opts) {
 	if (
 		opts &&
 		opts.flags &&
-		((opts.flags.get && opts.flags.get('frontPageAlpha')) || opts.flags.frontPageAlpha)
+		((opts.flags.get && opts.flags.get('betaHomePage')) || opts.flags.betaHomePage)
```
umbobabo, line 7: I thought we decided to go with `betaFrontPageOptIn`, didn't we? This is fine but I just need to know cause it's affect [my PR](https://github.com/Financial-Times/next-discovery-pages/pull/50)

umbobabo (approved):
Approved but with one clarification needed

myft/ui/navigationBetaTest.js
gyss: I'm sorry, I thought you read this slack message yesterday: 

https://financialtimes.slack.com/archives/C010BC61NDS/p1603293205169100?thread_ts=1603291188.165000&cid=C010BC61NDS

Luke told me we'll probably that OptIn env var in the future: https://github.com/Financial-Times/n-myft-ui/pull/428

gyss: see above comment from Luke

PR #432, opened 2021-02-15
fix: pin to date-fns version that do no use TS

umbobabo:
date-fns is migrating to TS, they are doing this progressively without transpiling the code


fenglish (approved):
👍

PR #436, opened 2021-04-12
upgrade to node v12

gyss:
Publishing to Snyk is not working.

This upgrade to node v12 will help with the Snyk integration


taraojo (approved):
Looks good, I guess this got missed when [everything was moved](https://financialtimes.slack.com/archives/C041V9QA7/p1604058131070600) to v12

gyss:
Probably this was missed since it was an only Bower package, but it's still using node to to the build

PR #438, opened 2021-04-16
Remove myft header tooltip

fenglish:
[Jira ticket](https://financialtimes.atlassian.net/browse/CON-853)
This feature was added as an AB test and the result was not successful.
This commit is to revert this commit 1c75cb0.
[This is the PR to add the feature](https://github.com/Financial-Times/n-myft-ui/pull/317)


Lily2point0 (approved):
Looks good

PR #443, opened 2021-08-03
CON-1097 Add prefix text to aria-label

chivchila:
Part 1. 
The prefix text for area-label will be feeding in from next-myft-page (Part 2) once this library is released


Lily2point0 (approved):

gyss:
looks good! just a small note: could you squash into 1 commit?

PR #450, opened 2021-12-13
Ci 868 follow button jsx

noelenwenede:
As part of the migration of `next-article` from `handle-bars` to `jsx`, all `handlebars HTML` files in shared libraries owned and consumed by `next-article` are being migrated to `jsx`.
 
This Pull request is the `unofficial` beginning of the migration of `myft-ui` from `handlebars` to `jsx`(essentially because no `jsx` files exist in this repository). The PR focuses on the `follow-button.html` partial and all child partials it consumes.

The changes consist of:
- the initial set-up for rendering jsx components on the server, 
- a `demo.jsx` to render all migrated components, 
- follow-button.jsx component.
- csrf-token/input.jsx component

NB: I have rendered the jsx component manually using a combination of `PageKitReactJSX` and `Handlebars`. This was done to reuse the `demo-layout.html` file.


AliceArmstrong (commented):
I don't think I'll have time to do a full review today, but I've done a couple of picky formatting suggestions (sorry 😂).

AliceArmstrong (commented):
Got a question from before the break that I apparently never submitted.

components/csrf-token/input.jsx
```diff
@@ -0,0 +1,24 @@
...
+		inputProps = {
+			...inputProps,
+			csrfToken
+		};
+	}
+
+	return (
+		<input
+			data-myft-csrf-token
+			{...inputProps}
```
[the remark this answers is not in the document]
noelenwenede: That's been added now

components/follow-button/follow-button.jsx
```diff
@@ -0,0 +1,174 @@
...
+		flags,
+		extraClasses,
+		conceptId,
+		variant,
+	} = props;
+
+	const formProps = generateFormProps(props);
+	const buttonProps = generateButtonProps(props);
+
+	const getVariantClass = (variant) => variant? `n-myft-follow-button--${variant}`: '';
```
AliceArmstrong, line 148: **Opinion:** I think that spacing out ternaries make them a bit more readable.
```suggestion
	const getVariantClass = (variant) => variant ? `n-myft-follow-button--${variant}` : '';
```

AliceArmstrong (commented):
Nice work! I've left a small suggestion, but it's more opinion-based than anything.

AliceArmstrong (approved):
🏅

demos/app.js
```diff
@@ -1,9 +1,17 @@
+require('sucrase/register');
 const express = require('@financial-times/n-internal-tool');
 const chalk = require('chalk');
 const errorHighlight = chalk.bold.red;
 const highlight = chalk.bold.green;
-
+const { PageKitReactJSX } = require('@financial-times/dotcom-server-react-jsx');
+let fs = require('fs');
```
gyss, line 7: not a biggie, but is there a reason to use `let` instead of `const` here?

noelenwenede: Happy to change it to `const`.

gyss (approved):

demos/app.js
noelenwenede, line 7:
```suggestion
const fs = require('fs');
```

PR #451, opened 2022-01-11
added rollup to bundle jsx components into js bundles.

noelenwenede:
The PR is a follow up from https://github.com/Financial-Times/n-myft-ui/pull/450 which was the initial set up for the migration of handlebars partials to jsx.

This change focuses on compiling the jsx templates into javascript.

As part of this PR the following changes were made:
- rollup.config.js
- An index.js file was created to barrel export all jsx components for rollup.

NB: The publish script currently publishes all contents of the repository so the build/compile step for jsx with roll up doesn't bundle it's CSS files as well.


jkerr321 (approved):

umbobabo (commented):
Question: With the introduction of rollup wouldn't we have 2 bundlers (Rollup and Webpack via DotCom)?

I'm asking because this PR introduces 8 new packages that then will need to be maintained.

noelenwenede:
> Yes I agree with Umberto - we should look at using standardised tooling for this where possible.
> 
> **Question:** what is the build step used for? Is it only used for the demos? or is it required for the consuming app to use this package?

@GlynnPhillips the `@sucrase/register` plugin we use for transpiling `jsx` to js in our apps doesn't transform `jsx` files from `node_modules`. Sadly they don't provide a config option to ignore specific files or paths so the compiled version of the component is required for consumers of the library using `@sucrase/register`.

PR #453, opened 2022-01-21
CON-1549 - migration to npm7

gyss:
Migration to NPM7


.circleci/config.yml
```diff
@@ -2,10 +2,10 @@
...
 
 references:
-  container_config_node12: &container_config_node12
+  container_config_node: &container_config_node
     working_directory: ~/project/build
     docker:
-      - image: circleci/node:12-browsers
+      - image: circleci/node:14-browsers
 
   container_config_lambda_node12: &container_config_lambda_node12
```
gyss, line 10: to be confirmed: is this container needed?

umbobabo: If you search this file for the reference `container_config_lambda_node12` and the reference is not present this lines should not be necessary cause the container is not used.

.circleci/config.yml
```diff
@@ -20,13 +20,13 @@ references:
 
   npm_cache_keys: &npm_cache_keys
     keys:
-      - v3-dependency-npm-{{ checksum "package.json" }}-
-      - v3-dependency-npm-{{ checksum "package.json" }}
-      - v3-dependency-npm-
+      - v4-dependency-npm-{{ checksum "package-lock.json" }}-
```
umbobabo, line 23: It's fine to bump the version, however have changed this part to use the package-lock instead of the package, the checksum will have a different value so the cache will be invalidated anyway.

package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "commit": "commit-wizard",
-    "prepare": "npx snyk protect || npx snyk protect -d || true"
+    "prepare": "npx snyk protect || npx snyk protect -d || true",
+    "preinstall": "check-engine"
```
umbobabo, line 10: Nitpick: I believe this part should be like [here](https://github.com/Financial-Times/n-topic-card/commit/250b8210ae809f71e51ac8f46921e9dbf0facc8a#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R33) or you can have trouble on CI because the packages are deleted.

umbobabo (approved):

PR #473, opened 2022-02-21
V25

noelenwenede:
The PR is a follow up from https://github.com/Financial-Times/n-myft-ui/pull/450 and a rewrite of https://github.com/Financial-Times/n-myft-ui/pull/451 which was the initial set up for the migration of handlebars partials to jsx.

This change focuses on compiling the jsx templates into javascript using webpack. The transpiling step is required to avoid the consuming applications transpiling packages from node_modules which could potentially cause issues with `sucrase` (the JIT transpiler used across FT) and the `handlebars renderReactComponent` helpers.

There exist a cross dependency between some handlebars partials and jsx components. This cross dependencies have caused failures in how the partials and jsx components are used. `Due to the previous stated reason, these migrated components cannot be released individually`. To fix this issue, all partials that depend on the migrated jsx components have been migrated as part of the webpack build feature.

This PR includes migration for the following handlebars partial:
- ConceptList
- Collections
- SaveForLater
- PinButton
- InstantAlert

It is easier to review on a commit basis. The first commit adds webpack config for transpiling the jsx components. 
Other commits have the following structure in specific order:
- Add component jsx
- Remove handlebars partial.

I'll be happy to look through the PR with any reviewer.


components/save-for-later/save-for-later.jsx
```diff
@@ -0,0 +1,103 @@
...
+		let props = {
+			type: 'submit',
+			'data-trackable': trackableId ? trackableId : 'save-for-later',
+			'data-text-variant': appIsStreamPage !== true ? 'save-button-with-icon-copy' : 'save-button-longer-copy',
+			'data-content-id': contentId,
+			className: saveButtonWithIcon ? 'n-myft-ui__save-button-with-icon' : `n-myft-ui__button ${variant ? `n-myft-ui__button--${variant}` : ''}`
+		};
+
+		if (isSaved) {
+			let titleText = `${title ? `${title} is` : ''} Saved to myFT`;
```
AliceArmstrong, line 54: **Suggestion:** I know this is the grammar of the old component, but I think it would be more correct to change 'Saved' to 'saved' (Same in line 63). However, I had a look for where it's used to show to Olga, and I actually couldn't find it on the site so I wonder if it's even in use.

components/pin-button/pin-button.jsx
```diff
@@ -0,0 +1,40 @@
...
+								data-pressed-text={`${name} pinned in myFT.`}
+								data-unpressed-text={`Unpinned ${name} from myFT.`}
+							></div>
+							<button id={`myft-pin-button__${id}`}
+								className="myft-pin-button"
+								data-prioritise-button
+								data-trackable="prioritised"
+								data-concept-id={id}
+								data-prioritised={prioritised ? true : false}
+								aria-label={`${prioritised ? 'Unpin' : 'Pin'} ${name} ${prioritised ? 'from' : 'in'} myFT`}
```
AliceArmstrong, line 29: **Comment:** I noticed that some aria-labels in other components say 'my F T' instead of 'myFT', and I was wondering if this is so that screen readers read it better. It might be fine, but I'm not sure how a screen reader would handle reading out 'myFT' to a user.

components/csrf-token/input.jsx
```diff
@@ -5,18 +5,12 @@ export default function CsrfToken ({ cacheablePersonalisedUrl, csrfToken }) {
...
-	}
-
-	if(csrfToken) {
 		inputProps.value = csrfToken;
 	}
 
 	return (
 		<input
+			data-myft-csrf-token
 			{...inputProps}
```
AliceArmstrong, line 14: **Question:** I might be a bit confused and reading this wrong, but wouldn't this make `data-myft-csrf-token` and `{...inputProps}` two separate attributes when before it was one attribute?

components/instant-alert/instant-alert.jsx
```diff
@@ -0,0 +1,73 @@
...
+			name: '_rel.instant',
+			className: `n-myft-ui__button n-myft-ui__button--instant n-myft-ui__button--instant-light${variant ? ` n-myft-ui__button--${variant}` : ''}${size ? ` n-myft-ui__button--${size}` : ''}`
+		};
+		return buttonProps;
+	}
+
+	return (
+		<Fragment>
+			{flags.myFtApiWrite &&
+				<form className={`n-myft-ui n-myft-ui--instant${hideButtonText ? ' n-myft-ui--instant--hide-text' : ''}${extraClasses? ` ${extraClasses}`: ''}`}
```
AliceArmstrong, line 59:
```suggestion
				<form className={`n-myft-ui n-myft-ui--instant${hideButtonText ? ' n-myft-ui--instant--hide-text' : ''}${extraClasses ? ` ${extraClasses}` : ''}`}
```

AliceArmstrong (commented):
Left some minor comments.

components/concept-list/concept-list.jsx
```diff
@@ -0,0 +1,63 @@
...
+import FollowButton from '../follow-button/follow-button';
+
+export default function ConceptList ({ flags, concepts, contentType, conceptListTitle, trackable, csrfToken, cacheablePersonalisedUrl }) {
+
+	const {
+		myFtApi,
+		myFtApiWrite
+	} = flags;
+
+	const generateTrackableProps = (primary, seconday) => {
```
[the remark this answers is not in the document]
noelenwenede: Nice spot, it wasn't deliberate. Thanks

components/csrf-token/input.jsx
noelenwenede: There was an error with the initial migration. This change fixed it. See https://github.com/Financial-Times/n-myft-ui/blob/bf2145fcfd211001dfab6a3a271dd72808d062bd/components/csrf-token/input.html for comparism.

components/save-for-later/save-for-later.jsx
```diff
@@ -0,0 +1,103 @@
...
+						}
+
+						{
+							appIsStreamPage === true && <span>{isSaved ? 'Saved' : 'Save'}</span>
+						}
+					</Fragment>
+				}
+			</Fragment>
+		}
+	</Fragment>);
```
boyewale, line 38: A lot of Fragment use in this file

components/pin-button/pin-button.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { Fragment } from 'react';
+import CsrfToken from '../csrf-token/input';
+export default function PinButton ({ showPrioritiseButton, id, name, directType, prioritised, csrfToken, cacheablePersonalisedUrl }) {
+
+	const getAction = () => `/__myft/api/core/prioritised/concept/${id}?method=${prioritised ? 'delete' : 'put'}`
+
+	return (
+		<Fragment>
+			{showPrioritiseButton &&
+				<Fragment>
```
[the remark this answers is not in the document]
noelenwenede: The version of webpack used in this repo doesn't support the `react preset` that has support for shorthand `Fragments <></>`. Upgrading the version of webpack would require more changes to logic that are not part of this migration so I opted to use Fragments as part of this migration. It can be changed when webpack is upgraded.

boyewale (approved):

PR #480, opened 2022-03-15
Switch to new NPM publish token

lbarrous:
Why?
We're in the process of looking at making our NPM publish process more secure. As part of that, we want to migrate our existing NPM publish jobs to use a new publish token, then revoke all of the old tokens.

@wheresrhys suggested using this package as a test to check that it works - for more info on the overall plan, see [this draft doc](https://docs.google.com/document/d/1xnUfCgUVvfotBSWFVD1ajpVy9qhnTc8qfFvkY6fN5Gg/edit#).

What?
Use the NPM_AUTH_TOKEN variable within the npm-publish-token context.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,50 +2,60 @@
 # template: component
 
 references:
-  container_config_node: &container_config_node
+  container_config_node:
+    &container_config_node
     working_directory: ~/project/build
     docker:
       - image: circleci/node:12-browsers
 
   workspace_root: &workspace_root ~/project
 
-  attach_workspace: &attach_workspace
+  attach_workspace:
+    &attach_workspace
     attach_workspace:
       at: *workspace_root
 
-  npm_cache_keys: &npm_cache_keys
+  npm_cache_keys:
+    &npm_cache_keys
     keys:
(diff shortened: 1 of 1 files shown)


juancancela (approved):
LGTM!

.circleci/config.yml
```diff
@@ -62,11 +72,13 @@ jobs:
     <<: *container_config_node
     steps:
       - checkout
-      - node/install-npm:
+      - node/install-npm: null
```
umbobabo, line 75: Question: is this change expected?
```suggestion
      - node/install-npm: null
```

fenglish: @lbarrous Please ask about this is expected or not to Platform team
https://financialtimes.slack.com/archives/C010BC61NDS/p1647438855529749?thread_ts=1647336271.361169&cid=C010BC61NDS

serena97: good spot - sorry this change isn't expected, we will take a look into this. once you remove this change in line 75 it is ready to merge

fenglish: Thank you @serena97.
@lbarrous Could you remove this change please?

fenglish (approved):
@lbarrous Looks good. Thank you for the change

PR #485, opened 2022-04-08
UG-815: Add error reporting through Sentry

meel-io:
## Description

Initialises `o-errors`on the save article variant to report errors to `next-article`'s Sentry project.
Adds `try/catch` structures to the `showSaveArticleToListVariant` function and all its components' event listeners. These structures catch the errors and enrich them with component property with value `professorLis`t. We'll use this property to create a custom alert for errors coming from the variant

## How to test this code

- Check out this branch locally and run: 
  - `npm link`
  - `bower link`
- Check out the branch `feature/UG-815-initialise-o-errors-for-professorlists` on `next-article` and run:
  - `npm link @financial-times/n-myft-ui`
  - `bower link n-myft-ui`
- Build and run `next-article`
  - `make build run`
- Go to FT Toggler and select the variant on the [professorLists flag](https://toggler.ft.com/#professorLists)
- Go to [myFT section](https://www.ft.com/myft/lists) and make sure you don't have any lists created. 
- Force an error on `openSaveArticleToListVariant` or on any of the event listeners. One way to force errors is to add a custom `throw new Error(message)` statement
- Go to [an article](https://local.ft.com:5050/content/aa9caaad-c19f-4451-b13b-e514b9c284a5), open the save article variant and trigger the error
- Inspect the Network tab on your Dev Tools and confirm that there are requests to Sentry endpoints

### DIFF
--- a/myft/ui/save-article-to-list-variant.js
+++ b/myft/ui/save-article-to-list-variant.js
@@ -3,36 +3,10 @@ import myFtClient from 'next-myft-client';
 import { uuid } from 'n-ui-foundations';
 import getToken from './lib/get-csrf-token';
 import oTooltip from 'o-tooltip';
-import oErrors from 'o-errors';
 
 const csrfToken = getToken();
 
 let lists;
-
-function initialiseErrorTracking () {
-	oErrors.init({
-		sentryEndpoint: 'https://d63d3b29fd374c3d95fa814a38c9bac9@o16684.ingest.sentry.io/79480',
-		transformError: function (data) {
-			const { context } = data;
-
-			if (context) {
-				const { extra = {} } = context;
-				const { info = {} } = extra;
-
-				if (info.component === 'professorLists') {
-					data.error.message = `[professorLists]: ${data.error.message}`;
-				}
-			}
-
-			return data;
-		},
-		filterError: function (data) {
-			const { error = { message: '' }, errormessage = '' } = data;
-
-			return errormessage.includes('[professorLists]') || error.message.includes('[professorLists]');
-		}
-	});
-}
 
 export default async function showSaveArticleToListVariant (name, contentId) {
 	try {
@@ -392,7 +366,6 @@ function triggerCreateListEvent (contentId) {
 }
 
 function handleError (error) {
-	initialiseErrorTracking();
 	document.body.dispatchEvent(new CustomEvent('oErrors.log', {
 		bubbles: true,
 		detail: {
--- a/myft/ui/save-article-to-list-variant.js
+++ b/myft/ui/save-article-to-list-variant.js
@@ -9,28 +9,30 @@ const csrfToken = getToken();
 
 let lists;
 
-oErrors.init({
-	sentryEndpoint: 'https://d63d3b29fd374c3d95fa814a38c9bac9@o16684.ingest.sentry.io/79480',
-	transformError: function (data) {
-		const { context } = data;
+function initialiseErrorTracking () {
+	oErrors.init({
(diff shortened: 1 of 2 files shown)


kavanagh:
@meel-io is there something specific in production that needs to be debugged with o-errors/Sentry? [We've not been using Sentry for client side errors for some time now](https://github.com/Financial-Times/dotcom-page-kit/blob/0b357c0fbc8f82ae19ba8812e9d2b855df2c4c10/docs/design-decisions/client-side-error-logging.md) so I'm wondering what it's needed for.

Is it possible to only download and initialise oErrors when it's needed? It looks like it will be initialised for all n-myft-ui clients (which is most of the site).

meel-io:
> @meel-io is there something specific in production that needs to be debugged with o-errors/Sentry? [We've not been using Sentry for client side errors for some time now](https://github.com/Financial-Times/dotcom-page-kit/blob/0b357c0fbc8f82ae19ba8812e9d2b855df2c4c10/docs/design-decisions/client-side-error-logging.md) so I'm wondering what it's needed for.
> 
> Is it possible to only download and initialise oErrors when it's needed? It looks like it will be initialised for all n-myft-ui clients (which is most of the site).

Thanks @kavanagh. I had the feeling we weren't using Sentry anymore, but I wasn't sure. Our team has been using Sentry for some specific use cases like this one, where we're pushing a lot of new client-side code and are concerned about introducing bugs. It should be temporal, though, and as soon as we know the test result and are confident that it didn't introduce any bugs, we'll remove it. 
We'll move the initialisation into the `showSaveArticleToListVariant` method, so it only gets called if the variant is active.
And knowing that Sentry isn't currently in use, I think we can disable the other alerts and add one that filters errors with the `professorLists` category, and have those alerts going to our team's support channel.

myft/ui/save-article-to-list-variant.js
```diff
@@ -3,12 +3,46 @@ import myFtClient from 'next-myft-client';
 import { uuid } from 'n-ui-foundations';
 import getToken from './lib/get-csrf-token';
 import oTooltip from 'o-tooltip';
+import oErrors from 'o-errors';
```
kavanagh, line 6: Thanks for moving the initialisation behind the flag. Could we do anything to only load oErrors library only when it's needed? I've got no idea how big it is but it's worth remembering that the myft-ui bundle is added to all pages so this is still adding an extra library download to home, stream and article pages.

For example you could load it as a standalone script from the origami build service.

jamesnicholls: Is it possible to move initialisation to `next-article`, and have `n-myft-ui` just trigger the events on document element (which I think is already happening)?

meel-io: Hi @kavanagh / @jamesnicholls. Great idea, thanks. We raised a [PR](https://github.com/Financial-Times/next-article) on next-article with those changes and updated this one to remove the o-errors initialisations

leannecornish-ft (approved):
LGTM - I was able to send a test to sentry 👍

kavanagh (approved):
Thanks for making the changes, this looks good to me know. Can I just confirm for the benefit of this PR that Sentry has been added to next-article?

meel-io:
> Thanks for making the changes, this looks good to me know. Can I just confirm for the benefit of this PR that Sentry has been added to next-article?

Thanks for the review and suggestions ⭐ Yes, those changes can be found on [this PR](https://github.com/Financial-Times/next-article/pull/4707)

PR #491, opened 2022-05-05
UG-874: Restore openFormHandler if list wasn't successfully created

meel-io:


jenniferemshepherd (approved):
works as described 👍

myft/ui/save-article-to-list-variant.js
```diff
@@ -112,6 +117,14 @@ async function openSaveArticleToListVariant (name, contentId) {
 		}
 	}
 
+	function scrollHandler () {
+		realignListener(createListOverlay.wrapper, window.scrollY);
+	}
+
+	function resizeHandler () {
+		realignListener(createListOverlay.wrapper);
```
kavanagh, line 125: Is it ok that the currentScrollPosition isn't updated here? Reading `realignOverlay` it seems like this could be a bug?

```
if(currentScrollPosition && Math.abs(currentScrollPosition - originalScrollPosition) < 120) {
    return;
}
originalScrollPosition = currentScrollPosition;
```

For example, is the state on the 3 invocation correct here? ...

```
const realignListener = realignOverlay(100)

realignListener(element, 250) // originalScrollPosition = 250 (because Math.abs(250 - 100) < 120 === false)
realignListener(element)      // originalScrollPosition = undefined
realignListener(element, 200) // originalScrollPosition = undefined (**is this expected?**)
```

myft/ui/save-article-to-list-variant.js
```diff
@@ -18,7 +18,8 @@ export default async function showSaveArticleToListVariant (name, contentId) {
 async function openSaveArticleToListVariant (name, contentId) {
 	function createList (list) {
 		if(!list) {
-			return;
+			if (!lists.length) attachDescription();
```
kavanagh, line 21: Will `lists` always be defined here?

meel-io: Right. It wasn't always defined. I added a new variable `haveLoadedLists` to determine if the lists have been loaded and declare `lists` with an initial array value.

PR #494, opened 2022-05-13
fix: CPORM-16 - Npm migration

JSRedondo:
Migrate from bower to npm.
Install components as peer dependencies with financial-times scope and fix references.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,5 +107,25 @@
       "pre-commit": "node_modules/.bin/secret-squirrel",
       "pre-push": "make verify -j3"
     }
+  },
+  "peerDependencies": {
+    "@financial-times/n-notification": "^8.2.2",
+    "@financial-times/o-buttons": "^7.5.0",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -107,5 +107,25 @@
       "pre-commit": "node_modules/.bin/secret-squirrel",
       "pre-push": "make verify -j3"
     }
+  },
+  "peerDependencies": {
+    "@financial-times/n-notification": "^8.2.2",
```
debugwand, line 112: these have bumped by a major version from the version in bower.json, so we don't know if there will be a breaking change. we also don't actually need to move the origami components from bower.json as part of this work

debugwand (changes requested):

package.json
JSRedondo: We can review the status of the task after your PR will be approved, just to make sure everything is ok.

JSRedondo:
Incompatible with current changes

PR #495, opened 2022-05-13
Update bower to npm packages

debugwand:
*  n-ui-foundations, next-myft-client and next-session-client moved from bower to npm
* added ftdomdelegate and superstore-sync directly instead of relying on n-ui-foundations
* dart sass replaces node sass so the demo builds
* remove prevent npm dependencies


JSRound (approved):
Perfect! Great the change from @node-sass to @sass

juancancela (approved):
LGMT!

PR #508, opened 2022-06-21
UG-870 Remove handleError method from save article to list variant

meel-io:
### Description

The `handleError` method was added to capture errors coming from the `save-article-to-list variant` component and post them through `o-errors`. 
We're removing it as part of rolling out the feature as we didn't find any issues during the test and `o-errors` initialisation [is being removed](https://github.com/Financial-Times/next-article/pull/4753) from `next-article` as well.

### DIFF
--- a/myft/ui/save-article-to-list-variant.js
+++ b/myft/ui/save-article-to-list-variant.js
@@ -7,11 +7,7 @@ const csrfToken = getToken();
 
 let lists;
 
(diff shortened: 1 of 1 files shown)


jenniferemshepherd:
actually, **question**: do we still need to wrap `openSaveArticleToListVariant` now?

jenniferemshepherd (approved):
thank you!

PR #517, opened 2022-07-22
ug-931 new list design: 

leannecornish-ft:
Jira ticket: [ug-931](https://financialtimes.atlassian.net/browse/UG-931)

This PR implements the following parts of the new list design (see [wireframe](https://www.figma.com/file/09mlE9M0XVG5fIZy99tls2/(Final)-Professor-Engagement?node-id=758%3A7466))

- The dialogue displays all the users existing lists, with the most recent one added at the top
- If the article is already contained within a list, the checkbox next to the list name is checked, otherwise it is unchecked

Screenshot:

Screenshot 2022-07-22 at 15 36 53

To test this code:

- Check out this branch and run npm link
- Clone next-article and run npm link @financial-times/n-myft-ui
- Update next-myft-client on next-article to v10.3.0
- Run next-article with npm start
- Switch on the manageArticleLists flag on [Toggler](https://toggler.ft.com/#manageArticleLists)
- Got to a local article, e.g. https://local.ft.com:5050/content/cf5d9a52-f193-4c7a-b5d0-4182b6ddcf3d
- Click on "Save" and the manageArticleLists modal should appear.
- Select 'add to a new list' and create a list. It should appear in the list of lists, checked.
- Repeat and check that the most recent one created is at the top
- Close the modal and reload the page. Click on the 'saved' button. The lists that you previously added the article should still be checked.


myft/ui/save-article-to-list-variant.js
```diff
@@ -318,12 +319,12 @@ function calculateLargerScreenHalf (target) {
...
-async function getLists () {
-	return myFtClient.getAll('created', 'list')
-		.then(lists => lists.filter(list => !list.isRedirect))
-		.then(lists => {
-			return lists.map(list => ({ name: list.name, uuid: list.uuid, checked: false }));
-		});
+async function getLists (contentId) {
+	return myFtClient.getListsContent()
+		.then(results => results.items.map(list => {
+			const isChecked = !!list.content && list.content.length && list.content.some(content => content.uuid === contentId);
```
jamesnicholls, line 325: **suggestion (non-blocking):**
```suggestion
			const isChecked = Array.isArray(list.content) && list.content.some(content => content.uuid === contentId);
```

This ensures that `list.content` is actually an array (as oppossed to possibly a string), and therefore contains the `some` method. Also it won't throw if `list.content` is `undefined`.

jamesnicholls (approved):
Nice work, the component works well. I've added one suggestion, but it's not a blocker.

I think the PR could use a rebase; I think some commits are now redundant. But again, this is not a blocker.

meel-io:
> Nice work, the component works well. I've added one suggestion, but it's not a blocker.
> 
> I think the PR could use a rebase; I think some commits are now redundant. But again, this is not a blocker.

Good point, thanks. I've committed the change. I'll rebase the PR too

fenglish (approved):
Looks good to me. If @jamesnicholls is happy with the changes after his approval, our side is ok with the changes 👍

PR #522, opened 2022-08-08
UG-837: Add Public toggle to new list form on manageArticleLists modal

meel-io:
### Description

- Adds the attribute `data-myft-ui-save-new-config` to the Save button. It'll use it to add configuration props to the `manageArticleLists` component
- Adds the value `showToggleProp` to that attribute if the `myftListPublicPrivateToggle` flag is on
- Passes a config object to the `manageArticleLists`. Passes the `name` and the `showToggleProp` properties through it.
- Conditionally shows the Public toggle on the `FormElement` if the `showToggleProp` is true
- Adds the markup for the "Public" toggle to the `FormElement`
- Captures the Public toggle state and sends it to the `createList` method. The `createList` method will use it to display a fake door message
- Changes the markup and style of the `FormElement` to add the design changes required for this feature

This toggle is part of a smoke test. For now, it doesn't change anything on the created list, but it'll help us measure interest in this feature

### How to test this code?

- Check out this branch and run `npm link`
- Clone `next-article` and  run `npm link @financial-times/n-myft-ui`
- Run `next-article` with `npm start`
- Switch on the `manageArticleLists` and `myftListPublicPrivateToggle` flags on [Toggler](https://toggler.ft.com/#manageArticleLists)
- Click on "Save" and the `manageArticleLists` modal should appear.
- Click on "Add to a new list" to display the form. The public toggle should appear below the text input. Changing its selection shouldn't have any effect on the list creation

| Desktop |Mobile |
| ------ | ----- |
|Screenshot 2022-08-10 at 12 59 35|Screenshot 2022-08-10 at 12 59 52|

Please make sure the changes are not impacting these consumer apps in unexpected ways:
 
- [x] [next-myft-page](https://github.com/Financial-Times/next-myft-page) ([Live](https://www.ft.com/myft), [Local](https://local.ft.com:5050/myft))
- [x] [next-video-page](https://github.com/Financial-Times/next-video-page) ([Live](https://www.ft.com/video), [Local](https://local.ft.com:5050/video))
- [x] [next-tour-page](https://github.com/Financial-Times/next-tour-page) ([Live](https://www.ft.com/tour), [Local](https://local.ft.com:5050/tour))
- [x] [next-search-page](https://github.com/Financial-Times/next-search-page) ([Live](https://www.ft.com/search), [Local](https://local.ft.com:5050/search))

### DIFF
--- a/components/save-for-later/save-for-later.html
+++ b/components/save-for-later/save-for-later.html
@@ -4,7 +4,8 @@
 	data-myft-ui="saved"
 	action="/myft/save/{{contentId}}"
 	data-js-action="/__myft/api/core/saved/content/{{contentId}}?method=put"
-	{{#if @root.flags.manageArticleLists}}data-myft-ui-save-new="manageArticleLists"{{/if}}>
+	{{#if @root.flags.manageArticleLists}}data-myft-ui-save-new="manageArticleLists"{{/if}}
+	{{#if @root.flags.manageArticleLists}}data-myft-ui-save-new-config="{{#if @root.flags.myftListPublicPrivateToggle}}showPublicToggle{{/if}}"{{/if}}>
 	{{> n-myft-ui/components/csrf-token/input}}
 	<div
 		class="n-myft-ui__announcement o-normalise-visually-hidden"
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -316,6 +316,11 @@ $spacing-unit: 20px;
 				justify-content: flex-end;
 				@include oTypographySans($scale: 2);
 			}
+
+			&-public {
+				width: calc(100% - 14px);
+				padding: 0 3px;
+			}
 		}
 
 		&-lists {
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -289,18 +289,32 @@ $spacing-unit: 20px;
 
 		&-form {
 			display: flex;
+			flex-direction: column;
 			width: calc(100% - 32px);
-			justify-content: space-between;
-			height: 40px;
-			gap: 8px;
-			padding: 0 16px 16px;
+			gap: 16px;
+			padding: 0 16px 12px;
 
 			& > * {
 				flex: 1 1 auto;
+				margin-bottom: 0;
 			}
 
 			.o-forms-input {
 				margin-top: 0;
+			}
+
+			&-toggle {
+				position: absolute;
+			}
+
+			&-toggle-label::after {
+				background-color: oColorsByName('white');
+			}
+
+			&-buttons {
+				display: flex;
+				justify-content: flex-end;
+				@include oTypographySans($scale: 2);
 			}
 		}
 
--- a/myft/main.scss
+++ b/myft/main.scss
@@ -239,7 +239,7 @@ $spacing-unit: 20px;
 		}
 
 		&-add-description {
-			margin: 4px 0;
+			margin: oSpacingByName('s1') 0;
 		}
 
 		&-heading {
@@ -258,7 +258,7 @@ $spacing-unit: 20px;
 
 		&-footer {
 			border-top: 1px solid oColorsByName('black-5');
-			padding: 16px;
+			padding: oSpacingByName('s4');
 		}
 
 		&-icon {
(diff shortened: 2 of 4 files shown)


myft/main.scss
```diff
@@ -289,19 +289,38 @@ $spacing-unit: 20px;
...
 		&-form {
 			display: flex;
+			flex-direction: column;
 			width: calc(100% - 32px);
-			justify-content: space-between;
-			height: 40px;
-			gap: 8px;
-			padding: 0 16px 16px;
+			gap: 16px;
+			padding: 0 16px 12px;
```
jamesnicholls, line 295: **suggestion (non-blocking):** These values are all related, correct? Could they be stored in a variable to make the relationship explicit?

myft/main.scss
```diff
@@ -289,19 +289,38 @@ $spacing-unit: 20px;
...
 				flex: 1 1 auto;
+				margin-bottom: 0;
 			}
 
 			.o-forms-input {
 				margin-top: 0;
 			}
+
+			&-toggle {
+				position: absolute;
```
jamesnicholls, line 307: **question:** Do you need to do this because it's in a flex container?

myft/main.scss
```diff
@@ -289,19 +289,38 @@ $spacing-unit: 20px;
...
+			}
+
+			&-toggle-label::after {
+				background-color: oColorsByName('white');
```
jamesnicholls, line 311: **suggestion:** Could you use the [internal brand](https://registry.origami.ft.com/components/o-forms@9.4.0/?brand=internal) instead of overridding the Origami CSS here?

myft/ui/save-article-to-list-variant.js
```diff
@@ -9,22 +9,24 @@ let lists = [];
...
+			if (!newList && !newList.name) attachDescription();
 			return contentElement.addEventListener('click', openFormHandler, { once: true });
 		}
 
-		myFtClient.add('user', null, 'created', 'list', uuid(), { name: list,	token: csrfToken })
+		myFtClient.add('user', null, 'created', 'list', uuid(), { name: newList.name,	token: csrfToken })
 			.then(detail => {
 				myFtClient.add('list', detail.subject, 'contained', 'content', contentId, { token: csrfToken }).then((createdList) => {
-					lists.unshift({ name: list, uuid: createdList.actorId, checked: true });
+					lists.unshift({ name: newList.name, uuid: createdList.actorId, checked: true, isShareable: !!newList.isShareable });
```
jamesnicholls, line 24: **question:** Does the `isShareable` list have any impact on the design? Is there any reason to store it on the list object?

jamesnicholls (commented):
The form looks great on desktop and mobile. I have some questions, just to understand the code better. And I have some concerns about the CSS overrides, but I'm not really sure if there is an alternative in this case.

myft/main.scss
meel-io: Good question. I tried using the `internal` brand but it was giving me errors. Maybe because [not all colors exist on all brands](https://registry.origami.ft.com/components/o-colors@5.3.0/readme/?brand=internal#palette-colours)?

myft/ui/save-article-to-list-variant.js
meel-io: I thought showing the fake door notification would depend on the result of the createLIst operation. We could pass it as an argument to the callback and then have the logic to show the notification there.
However, it's not necessary to pass it because it's already in the scope of the callback, but I thought it was more declarative that way.

myft/main.scss
meel-io: Yes. I'm trying to prevent the `input` tag from taking up any space so that the spacing only comes from the `gap` property

components/save-for-later/save-for-later.html
```diff
@@ -4,7 +4,8 @@
 	data-myft-ui="saved"
 	action="/myft/save/{{contentId}}"
 	data-js-action="/__myft/api/core/saved/content/{{contentId}}?method=put"
-	{{#if @root.flags.manageArticleLists}}data-myft-ui-save-new="manageArticleLists"{{/if}}>
+	{{#if @root.flags.manageArticleLists}}data-myft-ui-save-new="manageArticleLists"{{/if}}
+	{{#if @root.flags.manageArticleLists}}data-myft-ui-save-new-config="{{#if @root.flags.myftListPublicPrivateToggle}}showPublicToggle{{/if}}"{{/if}}>
```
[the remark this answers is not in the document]
meel-io: Yes, exactly. It doesn't add a lot of value for now, it only saves us from adding new attributes. But I imagined that, if we're releasing it to all users, we could have different variations of this modal in the future

myft/main.scss
jamesnicholls: Ah yeah, probably not.

jamesnicholls (approved):

PR #547, opened 2022-11-22
[CON-2137] Updated styles on myft button

juancancela:
**Description**

Updated myft button styles to make use of the [mix-blend-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode) property on the inverse theme to better calculate colors based on parent background color.

image

hover over
image

### DIFF
--- a/mixins/lozenge/_themes.scss
+++ b/mixins/lozenge/_themes.scss
@@ -42,7 +42,7 @@ $myft-lozenge-themes: (
 	)
 );
 
-@function getValue($key) {
+@function getThemeColor($key) {
 	@return map-get($theme-map, $key);
 }
 
--- a/mixins/lozenge/_themes.scss
+++ b/mixins/lozenge/_themes.scss
@@ -13,7 +13,8 @@ $myft-lozenge-themes: (
 		highlight: rgba(white, 0.8),
 		pressed-highlight: rgba(white, 0.2),
 		disabled: rgba(oColorsByName('white'), 0.5),
-		focus-outline: oColorsByName('white')
+		focus-outline: oColorsByName('white'),
+		mix-blend-mode: lighten
 	),
 	opinion: (
 		background: oColorsByName('oxford-40'),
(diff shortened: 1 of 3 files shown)


gyss:
could you please add some screenshots of how it was before?

mixins/lozenge/_themes.scss
```diff
@@ -41,7 +42,7 @@ $myft-lozenge-themes: (
 	)
 );
 
-@function getThemeColor($key) {
+@function getValue($key) {
```
gyss, line 45: I wonder if changing the name here could affect other projects that use `n-myft-ui`. Did you check if some other projects use this?

https://github.com/search?q=org%3AFinancial-Times+getThemeColor&type=code

Could you investigate for example if `x-dash` is using this function? 
https://github.com/Financial-Times/x-dash/blob/70a34125fb38456a5d4069c9b757eac3baeb0c3e/components/x-follow-button/src/styles/mixins/lozenge/_toggle-icon.scss

I guess that in order to avoid possible future problems maybe it's better to leave this function name unchanged

mixins/lozenge/_themes.scss
```diff
@@ -13,7 +13,8 @@ $myft-lozenge-themes: (
 		highlight: rgba(white, 0.8),
 		pressed-highlight: rgba(white, 0.2),
 		disabled: rgba(oColorsByName('white'), 0.5),
-		focus-outline: oColorsByName('white')
+		focus-outline: oColorsByName('white'),
+		mix-blend-mode: lighten
```
gyss, line 17: I guess this change is affecting as well next-stream-pages current production button. I wonder if the styles we want to change here are needed to deploy as well in next-stream-pages current styles (page not associated to dotcom pages)

juancancela: mmm, I'd say do. What do you think @kavanagh ?

kavanagh: If I recall, there are some hacks in next-stream-page for this button.

kavanagh (approved):

PR #554, opened 2022-12-08
CON-2091: Label in name, collections button (DAC)

camillecroci:
# The problem 

[Jira](https://financialtimes.atlassian.net/browse/CON-2091):
Change the order in which the topic of the button appear in label of collections. The aria label needs to sounds similar to the visual label, so have the extra info about the subject either before or after the "same string" as the one displayed on the button.

# Solution
Add the topic of the collection before the "Add add to myFT"

# How to test
When inspecting a collection, you should see the aria-label as such "Something collection: add to myFT"

### DIFF
--- a/components/collections/collections.html
+++ b/components/collections/collections.html
@@ -74,7 +74,7 @@
 							,
 						{{~/unless~}}
 					{{~/concepts~}}"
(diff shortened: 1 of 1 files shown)


fenglish (approved):
Thank you!

PR #555, opened 2022-12-08
CON-2075: Label in name of follow button in 'Added' state (DAC)

camillecroci:
# Problem
[Jira](https://financialtimes.atlassian.net/browse/CON-2075):  
The label in name criterion requires that the visible label (in this instance ‘added’) be contained in the accessible name (in this instance ‘remove world from myFT’). At the moment they are totally different: "Added" (visual) vs "Remove <Subject> from myFT". 

# Solution

Use the string "Added {{name}} to myFT: click to remove". Here the visual label 'Added' is the first in the aria label. The following information allows the visual impaired users to be aware of the action when clicking the button. 

# How to test
In the demo, you should find a button in the "Added" state. If you inspect it, you can find the aria-label that should say "Added {{name}} to myFT: click to remove".

Screenshot 2022-12-09 at 11 17 32

### DIFF
--- a/demos/templates/demo.html
+++ b/demos/templates/demo.html
@@ -37,6 +37,15 @@
 
 					<h2
 						class="demo-section__title">
+						Unfollow button
+					</h2>
+
+					{{#followButton}}
+						{{> n-myft-ui/components/follow-button/follow-button setFollowButtonStateToSelected=true }}
+					{{/followButton}}
+
+					<h2
+						class="demo-section__title">
 						x-dash follow button
 					</h2>
 
--- a/demos/templates/demo.html
+++ b/demos/templates/demo.html
@@ -26,31 +26,20 @@
 			class="demo-section">
 			<div class="o-grid-row">
 				<div data-o-grid-colspan="12">
-					<h2
-						class="demo-section__title">
-						Follow button
-					</h2>
-
 					{{#followButton}}
+						<h2
+							class="demo-section__title">
+							Follow button
+						</h2>
 						{{> n-myft-ui/components/follow-button/follow-button}}
 					{{/followButton}}
 
-					<h2
-						class="demo-section__title">
-						Unfollow button
-					</h2>
-
 					{{#followButton}}
+						<h2
+							class="demo-section__title">
+							Unfollow button
+						</h2>
 						{{> n-myft-ui/components/follow-button/follow-button setFollowButtonStateToSelected=true }}
-					{{/followButton}}
-
-					<h2
-						class="demo-section__title">
-						x-dash follow button
-					</h2>
-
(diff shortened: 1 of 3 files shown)


demos/templates/demo.html
```diff
@@ -35,6 +35,15 @@ <h1>{{title}}</h1>
...
 
+					<h2
+						class="demo-section__title">
+						Unfollow button
+					</h2>
+
+					{{#followButton}}
+						{{> n-myft-ui/components/follow-button/follow-button setFollowButtonStateToSelected=true }}
+					{{/followButton}}
+
```
fenglish, line 46: I think if you add a correct heading like other button (e.g. save buttons), it is nice. And please add demo screenshot in the description.

```
	{{#followButton}}
		<h2 class="demo-section__title">
			Follow button
		</h2>
		{{> n-myft-ui/components/follow-button/follow-button}}
	{{/followButton}}

	{{#followButton}}
		<h2 class="demo-section__title">
			Unfollow button
		</h2>
		{{> n-myft-ui/components/follow-button/follow-button setFollowButtonStateToSelected=true}}
	{{/followButton}}
```

camillecroci: I don't understand, it has a title that is the same as the one of the follow button (that was already there)?
Also i did not add a screenshot because the change is about 'aria-label' so it's not really visible. You are suggesting to add a screenshot of the inspector ?

fenglish: However you have added a new button with different state.

fenglish: Saved button does this in the demo
Screenshot 2022-12-09 at 11 01 12

fenglish: How about we do with the follow button the same like
Screenshot 2022-12-09 at 11 01 56

fenglish: Please move the title into the condition
```
{{#followButton}}
  <h2 class="demo-section__title">
   Unfollow button
  </h2>
  {{> n-myft-ui/components/follow-button/follow-button}}
 {{/followButton}}
```

camillecroci: should I do that for the existing follow button that was not in the condition as well ?

camillecroci: the x-dash follow button and the pin button titles are also not inside the handlebars tags. Should I modify those too?

fenglish: Yes, please. You will see it will much easy to find where the follow button's related code if we do that.

camillecroci: Ive added that, except for the pin button because there is a `{{each}}` helper which would create a title several times

camillecroci: Done :)

PR #560, opened 2023-01-16
bug fix: always add list element to dom so that new lists created within the component appear

leannecornish-ft:
# The change

Currently we conditionally add the list element to our modal if the users has existing lists when opening our modal. However, this means that, if a user goes on to add a list in the modal, the element is not in the DOM to re-render with the new list. 

This PR adds the element always, and conditionally hides it if there are no lists. It already gets unhidden as part of the submit add list process.

# How to test

1. Checkout the branch locally and run `npm link`
2. Locally in next-article, run `npm link @financial-times/n-myft-ui`, then build and run
3. In myft (prod or local), check that you have no lists. If you do and you want to keep them, I can provide a test user 🙂 
4. Visit a local article, e.g. https://local.ft.com:5050/content/f78fb12c-1b90-457d-86c7-efc4df6b0a23. Use the right hand myft bar to open the save article modal. Check that the experience is the same as prod, i.e.
Screenshot 2023-01-16 at 11 12 19
5. Add to a new list by filling in the details, and hit add. You should return to the initial modal view, with the new list present, e.g. 
Screenshot 2023-01-16 at 11 14 12
6. Add more lists and check that they appear.
7. Close the modal and reopen. The lists should appear, e.g. 
Screenshot 2023-01-16 at 11 15 11

### DIFF
--- a/myft/ui/save-article-to-list-variant.js
+++ b/myft/ui/save-article-to-list-variant.js
@@ -125,9 +125,10 @@ export default async function openSaveArticleToListVariant (contentId, options =
 	const resizeHandler = getResizeHandler(createListOverlay.wrapper);
 
 	createListOverlay.wrapper.addEventListener('oOverlay.ready', (data) => {
-		if (lists.length) {
-			const overlayContent = document.querySelector('.o-overlay__content');
-			overlayContent.insertAdjacentElement('afterbegin', listElement);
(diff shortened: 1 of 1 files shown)


jamesnicholls (approved):
Nice fix, works for me (once I sorted out the issue with the DOM delegate).

PR #561, opened 2023-01-23
ug-1103 - bug fix: move delegate creation inside init function 

leannecornish-ft:
### Description

We received reports of users regularly getting the core experience when interacting with the save article component. This PR aims to improve the situation by doing the following:

- Moves the DOM delegate into the init function of the buttons. This means that is should only be run after the DOM is ready (i.e. emitted the DOMContentLoaded event), as this is listened for before initialising components. 
- Checks the ready state of the DOM in case it still isn't ready, and adds an event listener to create the delegate if not.
- Moves this into a function that any of the buttons' init functions can use

However, it should be noted that people with exceptionally slow download speeds may still get the core experience as they will be able to interact with the SSR button before any client side javascript has had the chance to run.

### How to test

It's difficult to test that it fixes the bug, because it's very difficult to reproduce it! 

First try to run next-article locally and replicate the issue. I found it appears for me if I throttle my bandwidth to fast 3G in dev tools (making sure to wait until the page has loaded - this will catch the bug where the delegate has failed). Also make sure that you have disabled your cache.

1. Clone the branch locally and run `npm link`
2. In next-article run `npm link @financial-times/n-myft-ui`, build and run
3. With the same speed, test that the saved modal appears (once the page has fully loaded). 
4. Try will a slower speed/before the client-side javascript has loaded (or disable it), the core experience should still work

### Screenshots
Before:
Screenshot 2023-01-23 at 16 15 35

After:
Screenshot 2023-01-23 at 16 21 30

### DIFF
--- a/myft/ui/myft-buttons/init.js
+++ b/myft/ui/myft-buttons/init.js
@@ -8,9 +8,16 @@ import Delegate from 'ftdomdelegate';
 import personaliseLinks from '../personalise-links';
 import doFormSubmit from './do-form-submit';
 import enhanceActionUrls from './enhance-action-urls';
+import readyState from 'ready-state';
 
-const delegate = new Delegate(document.body);
 let initialised;
+
+function createDelegate (action, selector, cb) {
+	readyState.domready.then(() => {
+		const delegate = new Delegate(document.body);
+		delegate.on(action, selector, cb);
+	});
+}
 
 function getInteractionHandler (relationshipName) {
 	return (ev, formEl) => {
@@ -29,7 +36,7 @@ function anonEventListeners () {
 	};
 
 	['followed', 'saved'].forEach(action => {
-		delegate.on('submit', relationshipConfig[action].uiSelector, event => {
+		createDelegate('submit', relationshipConfig[action].uiSelector, event => {
 			event.preventDefault();
 			nNotification.show({
 				content: messages[action],
@@ -77,7 +84,7 @@ function signedInEventListeners () {
 					});
 				});
 
-			delegate.on('submit', uiSelector, getInteractionHandler(relationshipName));
+			createDelegate('submit', uiSelector, getInteractionHandler(relationshipName));
 		}
 	});
 }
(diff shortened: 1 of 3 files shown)


myft/ui/myft-buttons/init.js
```diff
@@ -9,9 +9,21 @@ import personaliseLinks from '../personalise-links';
...
+	let delegate;
+	if (document.readyState !== 'loading') {
+		delegate = new Delegate(document.body);
+		delegate.on(action, selector, cb);
+	} else {
+		window.addEventListener('DOMContentLoaded', function () {
+			delegate = new Delegate(document.body);
+			delegate.on(action, selector, cb);
+		});
+	};
```
jamesnicholls, line 24: **suggestion (non-blocking):** You could use the `ready-state` library here, although I don't think there is any practical difference, except that we use it across Customer Products already and it will save you a couple of lines of code.

myft/ui/myft-buttons/init.js
```diff
@@ -77,7 +89,7 @@ function signedInEventListeners () {
 					});
 				});
 
-			delegate.on('submit', uiSelector, getInteractionHandler(relationshipName));
+			createDelegate('submit', uiSelector, getInteractionHandler(relationshipName));
```
jamesnicholls, line 92: **question:** I don't think there is any practical difference, but you're now creating two delegates, whereas the previous code was just using one. Any reason for this change?

jamesnicholls (approved):
I have a suggestion and a question, but this works really well, thanks.

myft/ui/myft-buttons/init.js
leannecornish-ft: I think that only one will ever be used as you either have `opts.anonymous` and get the anon event listeners, or you don't and get the signed in ones. But maybe I am over simplifying that?

jamesnicholls: Ah yes, you're right.

PR #562, opened 2023-01-25
Apply new origami focus stylings to myFT buttons

fenglish:
Origami team has updated the focus styling across ft.com via `o-normalise` [[The origami new focus styles](https://origami.ft.com/blog/2022/12/14/focus-styles/)]. 

**It causes two focus stylings are applied to `n-myft-ui` buttons on the consumer apps.**
- One is from `n-myft-ui`'s focus styling
- One is from the latest `o-normalise`'s focus styling.

We need to apply only `o-normalise`'s focus styling and get rid of the own focus styling to keep consistency across ft.com and also fix accessibility issues.

To achieve that solution, we need to use `o-buttons` mixins to `myftLozenge`' s toggle buttons.

### next-myft-page's screenshots
|Before<br>(Two different focus stylings are applied) | After<br>(one focus styling from o-normalise via o-button)|
|---|---|
|Screenshot 2023-01-25 at 16 03 16|Screenshot 2023-01-25 at 16 02 55|

### n-myft-ui's screenshots
|Before|After|
|---|---|
|Screenshot 2023-01-25 at 17 02 01|Screenshot 2023-01-25 at 17 01 04|

-----

### This PR includes some other fixes

- **Remove unused --standard styles**
`$myft-lozenge-themes` has `standard` theme for the default style. When `@each $theme in map-keys($myft-lozenge-themes)` is run to create classes by theme, it creates `***--standard` styles. However the default style is already created without the modifier.
For example, `n-myft-follow-button` styles are created with the standard theme settings. We don't need to create `.n-myft-follow-button--standard`. `n-myft-concept-pill` is the same. It doesn't need to create `n-myft-concept-pill--standard`.

- **Fix save button with icon focus size**

|Before|After|
|---|---|
|Screenshot 2023-01-25 at 16 37 54|Screenshot 2023-01-25 at 16 36 30|

- **Add digest promo templates in demo**
Screenshot 2023-01-25 at 16 55 31


camillecroci (approved):
So, I try to read everything and maybe my questions are from my ignorance of myftlozenge but: 
I'm not sure where the decision to change the color of buttons on focus comes from? I don't see it in the origami new style link so I'm not 100% sure it comes from that update. (By change of color I mean the follow button from grey to pink for example. (I feel like there was a discussion about this but I don't remember very well..)).   


Also concept pills don't get focus ? In firefox that means they dont even get selected when using tab (but it was already the case before that change so it was not introduced here)


Apart from those questions that are probably related to my ignorance, I do not have any other remarks cause it all seems to be working smoothly so Ill just approve it :)

fenglish:
@camillecroci Thank you for reviewing the PR. The focus colour is come from o-buttons. It is not a part of new origami focus styling change however, to apply new focus styling, I needed change the follow buttons to use o-buttons mixins. Therefore the focus colour is changed. Jazmin has approved the focus colour change. 
As you noticed, the concept pills issue is not by this change. Therefore I leave it as it is. (I assume the issue is only occurred on demo not in production. however I'm not sure because I never seen concept pills are used in `monochrome` / `opinion` / `alphaville` theme)

PR #580, opened 2023-05-12
ug-1233 anon save copy

leannecornish-ft:
As part of US Growth's [free article demand test](https://financialtimes.atlassian.net/browse/UG-1191), we anticipate that most visitors will be anonymous or unsigned in registered. 

### In this PR:

1) We want to tailor the default notification that pops up when attempting to save an article, with these users in mind, for our test cohort only.

This is temporary and will be removed after the test has run.

### Screenshots:

Before:
Screenshot 2023-05-15 at 10 43 54

After - **for our test cohort only**:
Screenshot 2023-05-15 at 10 45 39

2) We also want to address a bug, whereby the notification is always rendered under the sticky header and is therefore not visible. This is a particular problem as the user has often scrolled and received the sticky header before attempting to save, so clicking the save button appears to have action.

### How to test:

1. Clone the branch locally, install and run npm link
2. Run next-article locally and link to this code, i.e. run npm link @financial-times/n-myft-ui (may need to use --force)
3. In an incognito/private window, turn the [podcastReferral](https://toggler.ft.com/#podcastReferral) flag on and locally visit a free article (contact me on slack if you need a url)
4. Click the save button and check that the copy matches [the design](https://www.figma.com/file/CZWwhlLmIrYe2eH83B1xu3/3rd-Party-%3EFt.com?type=design&node-id=431-4060&t=Rd0GBvhZ68q8deAg-0), and that the link (the sign is doesn't work locally as it needs fastly? to redirect to accounts.ft.com)
5. Do the same with the flag toggled off, and check that the current experience is displayed
6. Play around with scrolling to check that the notification is always shown

### DIFF
--- a/myft/ui/lists.scss
+++ b/myft/ui/lists.scss
@@ -9,6 +9,8 @@
 }
 
 .n-notification {
+	z-index: 105; //bring notifications in front of the sticky header
+
 	.myft-ui__icon {
 		@include oIconsContent('logo', oColorsByName('black-80'), 20);
 		margin: 0 3px -5px;
--- a/myft/ui/myft-buttons/init.js
+++ b/myft/ui/myft-buttons/init.js
@@ -26,13 +26,19 @@ function getInteractionHandler (relationshipName) {
 	};
 }
 
-function anonEventListeners () {
+function anonEventListeners (flags = {}) {
 	const currentPath = window.location.pathname;
 	const subscribeUrl = '/products?segID=400863&segmentID=190b4443-dc03-bd53-e79b-b4b6fbd04e64';
 	const signInLink = `/login${currentPath.length ? `?location=${currentPath}` : ''}`;
 	const messages = {
 		followed: `Please <a href="${subscribeUrl}" class="myft-ui-subscribe" data-trackable="Subscribe">subscribe</a> or <a href="${signInLink}" data-trackable="Sign In">sign in</a> to add this topic to myFT.`,
 		saved: `Please <a href="${subscribeUrl}" class="myft-ui-subscribe" data-trackable="Subscribe">subscribe</a> or <a href="${signInLink}" data-trackable="Sign In">sign in</a> to save this article.`
+	};
+
+	// 11/5/23 - US Growth test for Free Article Demand, see https://financialtimes.atlassian.net/browse/UG-1191
+	// This will be cleaned up after the test as part of https://financialtimes.atlassian.net/browse/UG-1221
+	if (flags.get('podcastReferral')) {
+		messages.saved = `<a href="/register/access?multistepRegForm=multistep" data-trackable="Register">Register</a> for free or  <a href="${signInLink}" data-trackable="Sign In">sign in</a> in to save this article`;
 	};
 
 	['followed', 'saved'].forEach(action => {
@@ -99,7 +105,7 @@ export default function (opts) {
 		enhanceActionUrls();
 
 		if (opts && opts.anonymous) {
-			anonEventListeners();
+			anonEventListeners(opts.flags);
 		} else {
 			signedInEventListeners();
 			personaliseLinks();


myft/ui/lists.scss
```diff
@@ -9,6 +9,8 @@
 }
 
 .n-notification {
+	z-index: 105;
```
jamesnicholls, line 12: **suggestion (non-blocking):** MIght be worth adding a comment to explain why you're doing this.

myft/ui/myft-buttons/init.js
```diff
@@ -35,6 +35,12 @@ function anonEventListeners () {
 		saved: `Please <a href="${subscribeUrl}" class="myft-ui-subscribe" data-trackable="Subscribe">subscribe</a> or <a href="${signInLink}" data-trackable="Sign In">sign in</a> to save this article.`
 	};
 
+	// 11/5/23 - US Growth test for Free Article Demand, see https://financialtimes.atlassian.net/browse/UG-1191
+	// This will be cleaned up after the test as part of https://financialtimes.atlassian.net/browse/UG-1221
+	if (flags.flags && flags.flags.podcastReferral) {
+		messages.saved = `<a href='/register/access?multistepRegForm=multistep' data-trackable="Register">Register</a> for free or  <a href="${signInLink}" data-trackable="Sign In">sign in</a> in to save this article`;
```
jamesnicholls, line 41:
```suggestion
		messages.saved = `<a href="/register/access?multistepRegForm=multistep" data-trackable="Register">Register</a> for free or  <a href="${signInLink}" data-trackable="Sign In">sign in</a> in to save this article`;
```

myft/ui/myft-buttons/init.js
```diff
@@ -35,6 +35,12 @@ function anonEventListeners () {
 		saved: `Please <a href="${subscribeUrl}" class="myft-ui-subscribe" data-trackable="Subscribe">subscribe</a> or <a href="${signInLink}" data-trackable="Sign In">sign in</a> to save this article.`
 	};
 
+	// 11/5/23 - US Growth test for Free Article Demand, see https://financialtimes.atlassian.net/browse/UG-1191
+	// This will be cleaned up after the test as part of https://financialtimes.atlassian.net/browse/UG-1221
+	if (flags.flags && flags.flags.podcastReferral) {
```
jamesnicholls, line 40: **question:** Why `flags.flags`?

jamesnicholls (approved):
Works great. I have a couple of minor suggestions, but nothing blocking.

myft/ui/myft-buttons/init.js
leannecornish-ft: Because the flags constant evaluates to an object with the key `flags` returned from `pageKitFlags.init()`, which then gets added to the `clientOptions` object that is used to initialise `myftUi.ui`.

kavanagh: `flags.get('podcastReferral')` is the usual way of referencing a flag client side. Is that not possible here?

camillecroci (approved):
Works as expected

myft/ui/myft-buttons/init.js
```diff
@@ -35,6 +35,12 @@ function anonEventListeners () {
 		saved: `Please <a href="${subscribeUrl}" class="myft-ui-subscribe" data-trackable="Subscribe">subscribe</a> or <a href="${signInLink}" data-trackable="Sign In">sign in</a> to save this article.`
 	};
 
+	// 11/5/23 - US Growth test for Free Article Demand, see https://financialtimes.atlassian.net/browse/UG-1191
+	// This will be cleaned up after the test as part of https://financialtimes.atlassian.net/browse/UG-1221
+	if (flags.get('podcastReferral')) {
```
kavanagh, line 40: It's possible for flags.get to be undefined here in the case where `opts.flags` is undefined.

myft/ui/lists.scss
```diff
@@ -9,6 +9,8 @@
 }
 
 .n-notification {
+	z-index: 105; //bring notifications in front of the sticky header
```
kavanagh, line 12: z-index with arbitrary number always stands out to me as a code smell and to expect bugs. But usually the root problem is having a UI design where too many elements have to overlap each other - i.e. overuse of popouts, popovers and flyovers!

And is it ok to do it for _all_ instances of an `.n-notification`?

leannecornish-ft: I agree, it's not ideal. We discussed this in USG and couldn't think of why you wouldn't want a notification in front. If it's hidden it may appear your event had no associated action. @kavanagh do you have any advice?

kavanagh: I don't. My comment was more meant as a warning to be on the look out for bugs and perhaps some extra testing...

PR #588, opened 2023-08-08
PT-234 Instant Alerts Onboarding Popup

asugar13:
### Description:
This PR implements the instant alerts onboarding pop up modal. It will be opened when clicked on the component implemented in this related PR (https://github.com/Financial-Times/n-myft-ui/pull/590)

For now, this is just a modal that matches designs. The modal has been added to the demo. We will be implementing the functionality to open/close the modal on a subsequent PR.

[JIRA](https://financialtimes.atlassian.net/jira/software/c/projects/PT/boards/1655?modal=detail&selectedIssue=PT-234)

Designs: https://www.figma.com/file/yyojM73dWDAF5BPgrMv7du/PT-219-To-increase-the-number-of-instant-alert-users?type=design&node-id=703-6064&mode=design&t=UvW4KTXefFngMMCs-0

Screenshots:

https://github.com/Financial-Times/n-myft-ui/assets/10453619/01df058f-e209-4d1f-8beb-c4761a889d55

### DIFF
--- a/components/jsx/preferences-modal/main.scss
+++ b/components/jsx/preferences-modal/main.scss
@@ -1,0 +1,42 @@
+@import '@financial-times/o-colors/main';
+@import '@financial-times/o-forms/main';
+@include oForms();
+
+.n-myft-ui__preferences-modal {
+	background-color: oColorsByName('white-80');
+	border-radius: 10px;
+	border: 2px solid oColorsByName('black-5');
+	width: 275px;
+}
+
+.n-myft-ui__preferences-modal__content {
+	position: relative;
+	margin: 20px 20px 24px;
+}
+
+.n-myft-ui__preferences-modal__checkbox__message {
+	@include oTypographySans($scale: 0, $weight: 'semibold', $style: 'normal');
+}
+
+.n-myft-ui__preferences-modal__text {
+	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
+	margin-bottom: 0;
+}
+
+.n-myft-ui__preferences-modal__remove-button {
+	margin-top: 20px;
+	height: 32px;
+	width: 100%;
+	border-radius: 20px;
+	border: 2px solid oColorsByName('black-20');
+	background-color: oColorsByName('white-80');
+	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
+	color: oColorsByName('black-80');
+}
+
+.n-myft-ui__preferences-modal__remove-button:hover,
+.n-myft-ui__preferences-modal__remove-button:focus {
+	border: 2px solid oColorsByName('claret-60');
+	background-color: oColorsMix($color: 'claret-60', $percentage: 20);
+	color: oColorsByName('claret-60');
+}
--- a/components/jsx/preferences-modal/preferences-modal.jsx
+++ b/components/jsx/preferences-modal/preferences-modal.jsx
@@ -1,0 +1,41 @@
+import React from 'react';
+
+/**
+ * @typedef {object} PreferencesProperties
+ * @property {string[]} currentPreferences
+ * Current alert preferences the user has set up.
+ * @property {object.<string, boolean>} flags
+ * 	FT.com feature flags
+ */
+
+/**
+ * Create a popup modal to manage myFT alert preferences
+ * @public
+ * @param {PreferencesProperties}
+ * @returns {React.ReactElement}
+*/
+export default function InstantAlertsPreferencesModal({ flags, currentPreferences }) {
+	if (!flags.myFtApiWrite) {
+		return null;
+	}
+
+	const formattedCurrentPreferences = currentPreferences.join(', ')
+	return (
+		<div className="n-myft-ui__preferences-modal">
+			<div className="n-myft-ui__preferences-modal__content">
+					<span className="o-forms-input o-forms-input--checkbox">
+						<label htmlFor="receive-instant-alerts">
+							<input id="receive-instant-alerts" type="checkbox" name="receive-instant-alerts" value="receive-instant-alerts"/>
+							<span className="o-forms-input__label n-myft-ui__preferences-modal__checkbox__message">
+								Get instant alerts for this topic
+							</span>
+						</label>
+					</span>
+
+				<p className="n-myft-ui__preferences-modal__text">Your alerts are currently: {formattedCurrentPreferences}.</p>
+				<a className="n-myft-ui__preferences-modal__text" href="/myft/alerts">Manage your preferences here</a>
+				<button className="n-myft-ui__preferences-modal__remove-button">Remove from myFT</button>
+			</div>
+		</div>
+	);
+};
--- a/demos/app.js
+++ b/demos/app.js
@@ -15,7 +15,8 @@ const fixtures = {
 	conceptList: require('./fixtures/concept-list'),
 	pinButton: require('./fixtures/pin-button'),
 	instantAlert: require('./fixtures/instant-alert'),
-	followPlusInstantAlerts: require('./fixtures/jsx/follow-plus-instant-alerts')
+	followPlusInstantAlerts: require('./fixtures/jsx/follow-plus-instant-alerts'),
+	instantAlertsPreferencesModal: require('./fixtures/jsx/preferences-modal'),
 };
 
 const app = module.exports = nExpress({
--- a/demos/fixtures/jsx/preferences-modal.json
+++ b/demos/fixtures/jsx/preferences-modal.json
@@ -1,0 +1,3 @@
+{
+    "currentPreferences": ["email", "desktop"]
+}
--- a/demos/src/demo.scss
+++ b/demos/src/demo.scss
@@ -9,6 +9,7 @@ $system-code: "n-myft-ui-demo";
 @import '../../components/follow-button/main';
 @import '../../components/jsx/follow-plus-instant-alerts/main';
 @import '../../components/onboarding-cta/main';
+@import '../../components/jsx/preferences-modal/main';
 
 @import '../../components/digest-promo/main';
 .n-myft-digest-promo {
--- a/demos/templates/demo.html
+++ b/demos/templates/demo.html
@@ -29,6 +29,16 @@
 
 				<div data-o-grid-colspan="12">
 					<h2>JSX components</h2>
+
+					{{#instantAlertsPreferencesModal}}
+						<h3
+							class="demo-section__title">
+							Instant alerts popup to manage preferences
+						</h3>
+
+						{{{renderReactComponent localPath="components/jsx/preferences-modal/preferences-modal.jsx" flags=@root.flags}}}
+					{{/instantAlertsPreferencesModal}}
+
 					{{#followPlusInstantAlerts}}
 						<h3
 							class="demo-section__title">


components/instant-alerts-onboarding/instant-alerts-onboarding.html
```diff
@@ -0,0 +1,23 @@
+{{#if @root.flags.myFtApiWrite}}
+	<div class="n-myft-ui__instant-alerts-onboarding">
+		<div class="n-myft-ui__instant-alerts-onboarding__content">
+			<form>
+				<span class="o-forms-input o-forms-input--checkbox n-myft-ui__instant-alerts-onboarding__checkbox">
+					<label for="receive-instant-alerts">
+						<input type="checkbox" name="receive-instant-alerts" value="receive-instant-alerts">
```
asugar13, line 7: **Note**: a11y tests are not passing because of this checkbox. I need to figure out a way to submit the form by clicking on the checkbox.

components/instant-alerts-onboarding/main.scss
```diff
@@ -0,0 +1,58 @@
...
+
+.n-myft-ui__instant-alerts-onboarding__list-preferences {
+	margin-bottom: 0;
+	font-size: 14px;
+}
+
+.n-myft-ui__instant-alerts-onboarding__manage-preferences {
+	margin-top: 0;
+	margin-bottom: 20px;
+	// font-size: 14px;
```
asugar13, line 32: **Note**: I left the `font-size` commented out to highlight the fact that there are a few design inconsistencies in the font choices.

components/instant-alerts-onboarding/instant-alerts-onboarding.html
GlynnPhillips: **Suggestion:** A possible suggestion here (I am still thinking if its the best way) would be to have hidden submit button within the form. Then when the user clicks on the checkbox we can submit the form via JS instead of using the submit button.

components/instant-alerts-onboarding/main.scss
GlynnPhillips: Can you provide more details about what you mean by this?

components/instant-alerts-onboarding/instant-alerts-onboarding.html
```diff
@@ -0,0 +1,23 @@
+{{#if @root.flags.myFtApiWrite}}
+	<div class="n-myft-ui__instant-alerts-onboarding">
```
GlynnPhillips, line 2: **Suggestion:** I wonder if our naming could be more generic / future proofed? Maybe something like `n-myft-ui__follow-menu` or `n-myft-ui__follow-manage` so that if we want to add other options into here we can?

components/instant-alerts-onboarding/main.scss
```diff
@@ -0,0 +1,58 @@
...
+	border-radius: 10px;
+	border: 2px solid oColorsByName('black-5');
+	width: 275px;
+}
+
+.n-myft-ui__instant-alerts-onboarding__content {
+	margin: 20px 20px 24px;
+}
+
+.n-myft-ui__instant-alerts-onboarding__checkbox {
```
GlynnPhillips, line 16: **Comment:** I am not sure if we should achieve this in this way. However to be sure I would like to run the demos and play around with it. I am in the middle of doing something in n-myft-ui so I need to find a good stopping point to do that

components/instant-alerts-onboarding/instant-alerts-onboarding.html
asugar13: will go with `n-myft-ui__preference-modal`

components/instant-alerts-onboarding/main.scss
asugar13: good call. moving the relative positioning to `n-myft-ui__preferences-modal__content`

components/instant-alerts-onboarding/instant-alerts-onboarding.html
asugar13: This is a good suggestion. However, since removing the form passes linting, that's what I've done for now. The main reason is that I think implementing this functionality is part of a separate task to make the API call to change the user's preferences. However, let me know: I could add a simple onClick listener without much functionality to it.

components/jsx/instant-alerts-preferences/instant-alerts-preferences.jsx
```diff
@@ -0,0 +1,44 @@
...
+ * Create a popup modal to manage myFT alert preferences
+ * @public
+ * @param {PreferencesProperties}
+ * @returns {React.ReactElement}
+*/
+export default function InstantAlertsPreferencesModal({ flags, currentPreferences }) {
+	//I think in practice this would not be needed 
+	if (!flags.myFtApiWrite) {
+		return null;
+	}
```
GlynnPhillips, line 21: **Comment:** Yeah probably not needed but I still think its worth having as its a simply bit of protection just incase someone loads this component without checking first

components/jsx/instant-alerts-preferences/instant-alerts-preferences.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+	//I've created a ticket to handle cases where the user has no preferences set
+	const formattedCurrentPreferences = currentPreferences.join(', ')
```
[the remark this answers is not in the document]
asugar13: sounds good. In this case I added functionality for the fields to be populated dynamically. However there are no designs for the case where there are no alerts set up for the user.

components/jsx/instant-alerts-preferences/main.scss
```diff
@@ -0,0 +1,54 @@
...
+	width: 100%;
+	border-radius: 20px;
+	border: 2px solid oColorsByName('black-20');
+	background-color: oColorsByName('white-80');
+}
+
+.n-myft-ui__preferences-modal__remove-button__text {
+	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
+	color: oColorsByName('black-80');
+}
```
GlynnPhillips, line 46: **Question:** Do we need this extra child element within the button? If we are only applying font styles then they can be applied directly to the button

components/jsx/instant-alerts-preferences/main.scss
```diff
@@ -0,0 +1,54 @@
...
+.n-myft-ui__preferences-modal__manage-preferences {
+	margin-top: 0;
+	margin-bottom: 20px;
+	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
+}
+
+.n-myft-ui__preferences-modal__remove-button {
+	display: flex;
+	align-items: center;
+	justify-content: center;
```
GlynnPhillips, line 35: **Comment:** I don't think we need this to be flex, especially if we remove the child element. The horizontal centering can be achieved with `text-align: center` and the vertical centering can be achieved by creating the height with equal padding

components/jsx/instant-alerts-preferences/instant-alerts-preferences.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+							<span class="o-forms-input__label n-myft-ui__preferences-modal__checkbox__message">
+								Get instant alerts for this topic
+							</span>
+						</label>
+					</span>
+
+				<p class="n-myft-ui__preferences-modal__list-preferences">Your alerts are currently: {formattedCurrentPreferences}.</p>
+				<p class="n-myft-ui__preferences-modal__manage-preferences">Manage your preferences <a>here</a></p>
+				<button class="n-myft-ui__preferences-modal__remove-button"><p class="n-myft-ui__preferences-modal__remove-button__text">Remove from myFT</p></button>
```
GlynnPhillips, line 40: **Change request:** Buttons can only contain inline elements and `p` is a block level element.

Also see my comment about the css because I think this child element is not required anyway

components/jsx/instant-alerts-preferences/main.scss
```diff
@@ -0,0 +1,54 @@
...
+	margin: 20px 20px 24px;
+}
+
+.n-myft-ui__preferences-modal__checkbox__message {
+	@include oTypographySans($scale: 0, $weight: 'semibold', $style: 'normal');
+}
+
+.n-myft-ui__preferences-modal__list-preferences {
+	margin-bottom: 0;
+	font-size: 14px;
```
GlynnPhillips, line 23: **Comment:** Should we use an origami mixin here to set the font size?

components/instant-alerts-preferences/instant-alerts-preferences.html
GlynnPhillips, line 1: Do we need a handlebars version?

components/jsx/instant-alerts-preferences/instant-alerts-preferences.jsx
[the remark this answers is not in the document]
GlynnPhillips: Yeah I just want to be careful to avoid big bang work. The ticket you are working on calls out that its not inscope to create any of the functionality within the modal. There is already a ticket PT-236 for doing the preference display work

components/jsx/instant-alerts-preferences/instant-alerts-preferences.jsx
GlynnPhillips, line 1: As we have called the classnames for this to be preference modal should we make the file names match?

components/instant-alerts-preferences/instant-alerts-preferences.html
asugar13: whoops! forgot to remove it

components/jsx/instant-alerts-preferences/instant-alerts-preferences.jsx
asugar13: I can do that but I felt like one of the reasons to change the classNames was that the original name was a bit wordy. I think the file name gives a bit of a more descriptive explanation of what the component is there for.

GlynnPhillips: > name was a bit wordy

That wasn't a concern for me, I have no issue with wordy names that are descriptive. My concern here is you have a filename that makes it sound very specific but then then code is aimed to make it sound generic. We should pick 1 or the other

asugar13: the original name was `instant-alerts-onboarding` (https://github.com/Financial-Times/n-myft-ui/pull/588#discussion_r1290094114) I will go with `instant-alerts-preferences` instead

GlynnPhillips: > instant-alerts-preferences

I personally think that this is confusing because we have other components that already exist on the site to deliver "instant alerts preferences"

>  I think the file name gives a bit of a more descriptive explanation of what the component is there for


I think the name we choice (preference modal ) is descriptive without being too specific. There is no point in our code being generic so the component can be extended if the filename is then overly specific.

Whatever choice you go with I would like to see the code and the file name reflect each other.

components/jsx/instant-alerts-preferences/main.scss
asugar13: I believe this was needed when there was a <p> inside the <button> but it's no longer needed:
```    
display: flex;
align-items: center;
justify-content: center;
```

components/jsx/preferences-modal/main.scss
```diff
@@ -0,0 +1,46 @@
...
+	height: 32px;
+	width: 100%;
+	border-radius: 20px;
+	border: 2px solid oColorsByName('black-20');
+	background-color: oColorsByName('white-80');
+	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
+	color: oColorsByName('black-80');
+}
+
+.n-myft-ui__preferences-modal__remove-button:hover {
```
GlynnPhillips, line 42: **Change request:** Here I think we need to account for `focus` state as well to allow for keyboard navigation

```suggestion
.n-myft-ui__preferences-modal__remove-button:hover,
.n-myft-ui__preferences-modal__remove-button:focus {
```

components/jsx/preferences-modal/preferences-modal.jsx
```diff
@@ -0,0 +1,42 @@
...
+								Get instant alerts for this topic
+							</span>
+						</label>
+					</span>
+
+				<p class="n-myft-ui__preferences-modal__list-preferences">Your alerts are currently: {formattedCurrentPreferences}.</p>
+				<p class="n-myft-ui__preferences-modal__manage-preferences">Manage your preferences <a href="/myft/alerts">here</a></p>
+				<button class="n-myft-ui__preferences-modal__remove-button">Remove from myFT</button>
+			</div>
+		</div>
```
GlynnPhillips, line 40: **Change request:** This is now jsx so all the `class` attributes in here need to be `className`

components/jsx/preferences-modal/preferences-modal.jsx
```diff
@@ -0,0 +1,42 @@
...
+		return null;
+	}
+
+	const formattedCurrentPreferences = currentPreferences.join(', ')
+	return (
+		<div class="n-myft-ui__preferences-modal">
+			<div class="n-myft-ui__preferences-modal__content">
+					<span class="o-forms-input o-forms-input--checkbox">
+						<label for="receive-instant-alerts">
+							<input type="checkbox" name="receive-instant-alerts" value="receive-instant-alerts"></input>
```
GlynnPhillips, line 28: **Nitpick:** `input` can be self closing here if you want 

```suggestion
							<input type="checkbox" name="receive-instant-alerts" value="receive-instant-alerts" />
```

components/jsx/preferences-modal/preferences-modal.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+							<span class="o-forms-input__label n-myft-ui__preferences-modal__checkbox__message">
+								Get instant alerts for this topic
+							</span>
+						</label>
+					</span>
+
+				<p class="n-myft-ui__preferences-modal__list-preferences">Your alerts are currently: {formattedCurrentPreferences}.</p>
+				<p class="n-myft-ui__preferences-modal__manage-preferences">Manage your preferences <a href="/myft/alerts">here</a></p>
```
GlynnPhillips, line 37: **Comment/Change request:** Maybe this can't be a change request until we speak to Design or Product but we should avoid "here" or "click here" type of links as they are bad for accessibility. For example if a user is use assistive tech to navigate via links in the page they wont know what "here" is as they wont have the "manage your preference" context.

I would suggest we do something like the below 
```suggestion
				<a href="/myft/alerts" className="n-myft-ui__preferences-modal__manage-preferences">Manage your preferences </a>
```

GlynnPhillips (changes requested):
A couple of minor improvements but its looking good

components/jsx/preferences-modal/preferences-modal.jsx
asugar13: true. it was self-closing at some point in the review. Can't remember why I added the closing tag.

components/jsx/preferences-modal/preferences-modal.jsx
asugar13: Instead of committing your suggestion, should I add it to this ticket to discuss with designer once he's back from time off? : https://financialtimes.atlassian.net/browse/PT-249

asugar13: I've gone ahead and made that change and updated the ticket so we can run it by Mark when he's back.

GlynnPhillips (approved):

gyss:
Looks good @asugar13 ! just one thing, on mobile, looks like the component is taking more width than the grid container. Is this being considered?

Screenshot 2023-08-16 at 11 39 01

components/jsx/preferences-modal/preferences-modal.jsx
```diff
@@ -0,0 +1,41 @@
...
+		return null;
+	}
+
+	const formattedCurrentPreferences = currentPreferences.join(', ')
+	return (
+		<div className="n-myft-ui__preferences-modal">
+			<div className="n-myft-ui__preferences-modal__content">
+					<span className="o-forms-input o-forms-input--checkbox">
+						<label htmlFor="receive-instant-alerts">
+							<input type="checkbox" name="receive-instant-alerts" value="receive-instant-alerts"/>
```
metart43, line 28: I might be wrong but I reckon the association between `<label>` and `<input>` is done through the `id` prop.

So, you just simply need to add `id` property to the `<input>` checkbox

```jsx
<input id="receive-instant-alerts" type="checkbox" name="receive-instant-alerts" value="receive-instant-alerts" />
```

components/jsx/preferences-modal/preferences-modal.jsx
```diff
@@ -0,0 +1,41 @@
...
+							<span className="o-forms-input__label n-myft-ui__preferences-modal__checkbox__message">
+								Get instant alerts for this topic
+							</span>
+						</label>
+					</span>
+
+				<p className="n-myft-ui__preferences-modal__text">Your alerts are currently: {formattedCurrentPreferences}.</p>
+				<a className="n-myft-ui__preferences-modal__text">Manage your preferences here</a>
```
metart43, line 36: Sorry, I am not sure why there is an `<a>` tag with a `href` property.


Without a valid `href`, this element could be confusing to screen readers and other assistive technologies. If it's meant to be interactive (like a button), it's better to use a `<button>` element or provide appropriate roles and ARIA attributes.

metart43 (changes requested):

components/jsx/preferences-modal/preferences-modal.jsx
asugar13: Sounds good, I'll add the `id`. It seems like it's a good accessibility practice

components/jsx/preferences-modal/preferences-modal.jsx
asugar13: Good spot! I think the href went away in a series of rebases I did.

components/jsx/preferences-modal/preferences-modal.jsx
GlynnPhillips: Just for future reference, if you have an implicit label & input relationship (eg. the label is wrapping the input) then you don't need the `for` & `id` relationship (explicit relationship) - however doing both is also not a problem

GlynnPhillips: Explicit relationships are better for accessibility but they still don't work perfectly if combined with an implicit relationship. However our styling via forms does rely on an implicit relationship so I don't think theres a way around that

asugar13 (commented):
@gyss , this is how the component looks like on mobile when being imported into other apps such as `next-article`: 
Screenshot 2023-08-16 at 13 53 12

gyss (approved):

metart43 (approved):

PR #589, opened 2023-08-08
JSX Components & demos

GlynnPhillips:
**Description**

Enable components to be built in JSX instead of handlebars and for the demos to still work

**Justification**

We (Profile team) need to add some new/variants of components for an experiment we plan to run to onboard users to instant alerts. Whilst this repo is entirely handlebars the wider estate within Customer Products is moving towards JSX, we therefore felt that it would make more sense for this work to be done in JSX.

I have made the changes to enable JSX components to be demos with an example component first as a talking point incase @Financial-Times/content-discovery are aware of something I have overlooked or have plans that this conflicts with.

### DIFF
--- a/components/jsx/example/example.jsx
+++ b/components/jsx/example/example.jsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+export default function Example({ title }) {
+	return (
+		<p>{title}</p>
+	);
+};
--- a/demos/app.js
+++ b/demos/app.js
@@ -1,3 +1,5 @@
+require('sucrase/register');
+
 const nExpress = require('@financial-times/n-express');
 const chalk = require('chalk');
 const errorHighlight = chalk.bold.red;
@@ -12,7 +14,8 @@ const fixtures = {
 	collections: require('./fixtures/collections'),
 	conceptList: require('./fixtures/concept-list'),
 	pinButton: require('./fixtures/pin-button'),
-	instantAlert: require('./fixtures/instant-alert')
+	instantAlert: require('./fixtures/instant-alert'),
+	exampleJsx: require('./fixtures/jsx/example')
 };
 
 const app = module.exports = nExpress({
--- a/demos/templates/demo.html
+++ b/demos/templates/demo.html
@@ -21,11 +21,31 @@
 	<div class="o-grid-container o-grid-container--snappy demo-container">
 		<h1>{{title}}</h1>
 
+
+		<section
+			id="example-component"
+			class="demo-section">
+			<div class="o-grid-row">
+
+				<div data-o-grid-colspan="12">
+					<h2>JSX components</h2>
+					{{#exampleJsx}}
+						<h3
+							class="demo-section__title">
+							Example component
+						</h3>
+
+						{{{renderReactComponent localPath="components/jsx/example/example.jsx"}}}
+					{{/exampleJsx}}
+				</div>
+			</div>
+		</section>
 		<section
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "css-loader": "^0.23.1",
     "denodeify": "^1.2.1",
     "eslint": "6.5.1",
+    "eslint-plugin-react": "^7.33.1",
```
asugar13, line 50: **Praise**: Good call on adding a linter

demos/app.js
```diff
@@ -1,3 +1,5 @@
+require('sucrase/register');
```
asugar13, line 1: **Question:** This is required for JSX to work on this repo right?

asugar13 (commented):
Nothing blocking me from approving this PR but I would let content disco approve it instead. Maybe they have considerations against jsx.

demos/templates/demo.html
```diff
@@ -21,66 +21,86 @@
...
 		<h1>{{title}}</h1>
 
+
+		<section
+			id="example-component"
+			class="demo-section">
+			<div class="o-grid-row">
+
+				<div data-o-grid-colspan="12">
+					<h2>JSX components</h2>
```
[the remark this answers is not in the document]
GlynnPhillips: I am not sure what you mean by this comment 😂 could you explain a little more please

demos/app.js
GlynnPhillips: Yes that right

metart43:
Hey @GlynnPhillips, I have no objections towards the JSX components. They are more familiar to me than the handlebar's templating language. 

My only question - was there a specific reason why you have chosen React 16 over 17/18?

GlynnPhillips:
@metart43 thats a great question that made me realise its not needed however I will start by answering the questions of why React 16.

I just ran `npm install react` which choose v16 - which is happening I believe to avoid conflicts with other dependencies such as pagekits `"@financial-times/dotcom-server-handlebars` which ships with React v16.

Realising this made me rethink whether I needed React as a dependency in this project and the answer is No. Anything that consumes the JSX components from this package will need to supply React in the same way that the demos supply react. So I ended up removing the react dependency

components/jsx/example/example.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
gyss, line 1: should `react` be a peerDependency?

GlynnPhillips:
> Just that I'm not sure if there would be any unexpected side effect (in this case adding sucrase)

Any side effects would only impact the demos (this will come back to bite me..) so I think low chance/impact even if it does

components/jsx/example/example.jsx
GlynnPhillips: Yes thats a great point, that would be the best way highlight this. I will change that

GlynnPhillips: Done

package.json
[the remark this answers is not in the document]
GlynnPhillips: I think (I am not an expert on this) this is handled by eslint-config-next 
https://github.com/Financial-Times/eslint-config-next/blob/main/index.js#L72-L88

asugar13 (approved):
Approving this PR but can you please link it to a ticket in JIRA?

GlynnPhillips:
Closing because https://github.com/Financial-Times/n-myft-ui/pull/590 replaces it

PR #590, opened 2023-08-10
Follow + Instant alerts management button

GlynnPhillips:
_This is a continuation of the JSX demo/components PR #589 but with a real component added_

### Description

This is now a 2 part PR which will be easier to follow along in the commits.

Part 1: Add a structure for JSX components and allow for them to be demoed
Part 2: Create a new follow + instant alerts button

I will focus on explaining part 2 in this description.

The Accounts (Profile) team are planning to release a new instant alerts onboarding feature with the aim of increasing instant alert users.

The new onboarding will appear when you click on the follow button in certain locations on FT.com. At this point the users will be presented with the option to set up instant alerts for that topic or unfollow the topic. The same presentation will also be shown to the user when they click to unfollow a topic. This turns the button into more of a "manage" experience than just an on/off experience. Hopefully the [figma designs](https://www.figma.com/file/yyojM73dWDAF5BPgrMv7du/PT-219-To-increase-the-number-of-instant-alert-users?type=design&node-id=703-6064&mode=design) will help to visualise this experience for you.

This PR contains a new follow button component that contains the variants needed to create this experience. That includes showing if the user has instant alerts on or off (bell icon) and adds a down arrow to hopefully indicate that the button reveals more when clicking on it.

### Screenshots

Screenshot 2023-08-10 at 16 41 14

### Jira
https://financialtimes.atlassian.net/browse/PT-238

### Notes

This is still missing some features of the regular follow button, for example the ability to add variants/themes. This was a deliberate choice to avoid this initial release from becoming too large. Those features will be added in later PR's probably when the need for them becomes relevant - eg. we want to place the button on themed toppers


components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+
```
asugar13, line 2: **Thought**: I just thought of an alternative name" `follow-plus-dropdown`. It sounds a bit more reusable and it's shorter. This is non-blocking. Lmk what you think

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @typedef {object} FollowProperties
+ * @property {string} conceptId
+ * 	The ID for the concept
+ * @property {string} name
+ * 	The user facing label for the concept
+ * @property {boolean} setFollowButtonStateToSelected
```
asugar13, line 9: **Praise**: Good call on adding JSDocs

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
```diff
@@ -0,0 +1,80 @@
...
+ * 	An indicator to state whether the button state should be set to selected on the server side
+ * @property {boolean} cacheablePersonalisedUrl
```
asugar13, line 11: **Question**: Could you explain the `cacheablePersonalisedUrl` a bit more? In the comments section here is fine for now. I don't think I've understood its purpose fully yet.

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Create a follow plus instant alerts button component
+ * @public
+ * @param {FollowProperties}
+ * @returns {React.ReactElement}
+*/
+
+export default function FollowPlusInstantAlerts({ conceptId, name, setFollowButtonStateToSelected, cacheablePersonalisedUrl, setInstantAlertsOn, flags }) {
+	if (!flags.myFtApiWrite) {
```
asugar13, line 27: **Comment**: This should suffice for now but should we start using [myFTInstantAlertsOnboarding](https://toggler.ft.com/#myFTInstantAlertsOnboarding)?

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
```diff
@@ -0,0 +1,80 @@
...
+		'data-js-action': `/__myft/api/core/followed/concept/${conceptId}?method=delete`
+	} : {
+		'action': `/myft/add/${conceptId}`,
+		'data-js-action': `/__myft/api/core/followed/concept/${conceptId}?method=put`
+	};
+
+	const dynamicButtonAttributes = setFollowButtonStateToSelected && cacheablePersonalisedUrl ? {
+		'aria-label': `Added ${name} to myFT: click to manage alert preferences or remove from myFT`,
+		'title': `Manage ${name} alert preferences or remove from myFT`,
+		'data-alternate-label': `Add to myFT: ${name}`,
```
asugar13, line 42: **Question**: If the button is selected are you sure about this `data-alternate-label` and `data-alternate-text`?

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
```diff
@@ -0,0 +1,80 @@
...
+		'aria-pressed': true,
+		'data-alternate-text': 'Add to myFT'
+	} : {
+		'aria-label': `Add ${name} to myFT`,
+		'title': `Add ${name} to myFT`,
+		'data-alternate-label': `Added ${name} to myFT: click to manage alert preferences or remove from myFT`,
```
asugar13, line 48: Question: If the button is NOT selected are you sure about this `data-alternate-label` and data-alternate-text?

demos/fixtures/jsx/follow-plus-instant-alerts.json
```diff
@@ -0,0 +1,3 @@
+{
+    "title": "This is an example jsx component"
```
asugar13, line 2: **Question**: Wouldn't we want to add some properties to the fixture? I also think title is not used?

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
GlynnPhillips: Interesting point as its obviously similar to the "make it more generic" point I am made on yours and I hadn't even thought about how I had named mine 😂 

The difference here is theres 2 differences to this button from the normal follow button.

1. The change of the Icon from the tick to the bell
2. The arrow drop down

I went with `plus instant alerts` because I feel this is more descriptive in this situation of what both of those changes are trying to achieve. Whilst I agree the drop down could be used for more in the future that is not instant alerts there would still be the bell change. So I think that if/when the time comes that the drop down does more we could revisit this to decide if it needs a better name

demos/templates/demo.html
```diff
@@ -21,66 +21,104 @@
...
+						{{{renderReactComponent localPath="components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx" concept=this flags=@root.flags}}}
+					{{/followPlusInstantAlerts}}
+
+					{{#followPlusInstantAlerts}}
+						<h3
+							class="demo-section__title">
+							Unfollow plus instant alerts button
+						</h3>
+
+						{{{renderReactComponent localPath="components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx" concept=this setFollowButtonStateToSelected=true cacheablePersonalisedUrl=true flags=@root.flags}}}
```
asugar13, line 47: **Comment**: I think `concept` is not a valid prop for the component.

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
GlynnPhillips: In most places on FT.com the follow buttons render initially in a unfollowed state, then when the client side javascript kicks in it changes the state of the button to followed if the user follows that topic.

Thats because the server rendered page needs to be cached so we can't have personalised states.

There are times however when we do know that the page is personalised to that user or theres no cache at all. In those cased the server rendered button has the ability to render in the correct state straight away. A good example of this is the myFT feed page. They load straight away in the "followed" state because we know they are followed and we know that page will only ever be seen by that user. This is a nicer experience on the feed page rather than the user seeing all the buttons flickering between states as the page loads.

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
GlynnPhillips: We still need the `myFtApiWrite` flag as this is a safety switch.

The [myFTInstantAlertsOnboarding](https://toggler.ft.com/#myFTInstantAlertsOnboarding) flag would go outside of this component because its an AB test. So it needs to be placed somewhere that it can decide whether to use this new button or the old one

demos/fixtures/jsx/follow-plus-instant-alerts.json
GlynnPhillips: Whoops, good catch. I forgot to update this from the when it was a demo component. I will update now

demos/templates/demo.html
GlynnPhillips: Good spot. I changed this at the last minute yesterday and I forgot to update this location. I will fix now

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
GlynnPhillips: Yes I am sure.. next question 😂 

The alternate text should explain the opposite state to that of the current button state. eg. if the button state is "added" the alternate text should talk about "adding"

GlynnPhillips: See the original follow button is similar https://github.com/Financial-Times/n-myft-ui/blob/main/components/follow-button/follow-button.html#L28-L42

components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx
GlynnPhillips: https://github.com/Financial-Times/n-myft-ui/pull/590/files/97be9e5a4893a9a150c0c81056dff95ae2bdef49#r1291060841

demos/templates/demo.html
GlynnPhillips: Fixed, thanks for catching that

demos/fixtures/jsx/follow-plus-instant-alerts.json
GlynnPhillips: Fixed, thanks for catching that

asugar13 (approved):
looks good to me

demos/templates/demo.html
```diff
@@ -21,66 +21,104 @@
...
+				<div data-o-grid-colspan="12">
+					<h2>JSX components</h2>
+					{{#followPlusInstantAlerts}}
+						<h3
+							class="demo-section__title">
+							Follow plus instant alerts button (same appearance as normal follow button)
+						</h3>
+
+						{{{renderReactComponent localPath="components/jsx/follow-plus-instant-alerts/follow-plus-instant-alerts.jsx" conceptId=this.conceptId name=this.name flags=@root.flags}}}
+					{{/followPlusInstantAlerts}}
```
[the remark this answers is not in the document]
GlynnPhillips: I agree with your thinking but I did it this way because its consistent with the rest of the file 🤷 So if we did thus I would do it for everything else not just this feature

PR #591, opened 2023-08-15
PT-239 Trigger preferences modal when add to myFT Button is clicked

asugar13:
### Description
This PR implements the interaction between the new add to myFT button and the preferences modal. 
    -The preferences modal is closed/open by clicking on the add to myFT button.
    -The preferences modal is closed if you click anywhere else on the screen.

[JIRA](https://financialtimes.atlassian.net/jira/software/c/projects/PT/boards/1655?modal=detail&selectedIssue=PT-239)

screengrab: 

https://github.com/Financial-Times/n-myft-ui/assets/10453619/63014b6f-0dbf-4693-b4a8-c72d8188a9b5

### DIFF
--- a/components/jsx/preferences-modal/main.scss
+++ b/components/jsx/preferences-modal/main.scss
@@ -9,6 +9,13 @@
 	width: 275px;
 }
 
+//this class needs to be active in prod but cannot be used in the demo as it the position absolute makes it render on top of other components
+.n-myft-ui__preferences-modal-live {
+	position: absolute;
+	margin-top: 50px;
+	z-index: 999;
+}
+
 .n-myft-ui__preferences-modal__content {
 	position: relative;
 	margin: 20px 20px 24px;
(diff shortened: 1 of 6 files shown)


myft/ui/myft-buttons/init.js
```diff
@@ -99,9 +101,11 @@ export default function (opts) {
 	if (!opts.anonymous) {
 		setTokens();
 	}
+	instantAlertsOnboarding();
```
asugar13, line 104: Decision choice: I've decided to put this here for now as it's the file that initialises all myFT buttons. Please let me know your thoughts, I could make it a bit more dependent on the `instantAlertsOnboarding` flag on toggler and put it in its own dedicated file and call it here: https://github.com/Financial-Times/n-myft-ui/blob/main/myft/ui/index.js

components/jsx/preferences-modal/main.scss
```diff
@@ -7,6 +7,9 @@
 	border-radius: 10px;
 	border: 2px solid oColorsByName('black-5');
 	width: 275px;
+	position: absolute; //is on top of other jsx components in the demo
```
asugar13, line 10: **Update:** I will add the absolute positioning in the JS file instead so that the demo still renders nicely.

asugar13: (wrap demo in relative position)

asugar13: I still need to look into this

PR #594, opened 2023-08-31
Open close preference modal

GlynnPhillips:
### Description

This enables the preference modal to be toggled open and closed via a custom event. It also sets up the new follow + instant alert button to dispatch that event so the components work together.

### Whats included

1. Ability to open and close the preference modal
2. The follow + instant alerts button to act as a trigger for that open and close
3. Disabling the 1 click remove follow user journey that would normally happen when you click on a added state button
4. Now if the button is in added state and you click on it then it will only trigger preference modal and do nothing else
5. Very basic positioning logic to ensure the modal is always within the viewport

### Whats not included

Stuff that will happen in future PR's (I didn't want this to get too large)

1. Any functionality within the preference modal including the ability to remove the topic follow
2. Tracking
3. The ability to close the preference modal by clicking anywhere else on the page
5. Probably some stuff I have forgotten :)

### Screen recording

https://github.com/Financial-Times/n-myft-ui/assets/524573/bf5d8193-01f3-4893-9037-dcf45fc0584a

### Jira

https://financialtimes.atlassian.net/browse/PT-239

### DIFF
--- a/components/jsx/follow-plus-instant-alerts/main.scss
+++ b/components/jsx/follow-plus-instant-alerts/main.scss
@@ -32,3 +32,15 @@
 	background-position: 50% 41%;
 	margin-left: 12px;
 }
+
+.n-myft-follow-button--instant-alerts--open[aria-pressed=true]::after {
+	content: "";
+	@include oIconsContent(
+		$icon-name: 'arrow-up',
+		$color: oColorsByName('white'),
+		$size: 10
+	);
+	background-size: 21px;
+	background-position: 50% 41%;
+	margin-left: 12px;
+}
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -1,0 +1,23 @@
+const preferenceModalShowAndHide = ({ event, preferencesModal }) => {
+	if (!event.target) {
+		return;
+	}
+
+	preferencesModal.classList.toggle('n-myft-ui__preferences-modal--show');
+};
+
+export default () => {
+	/**
+	 * This feature is part of a test
+	 * Therefore we have built it to work within the known parameters of that test
+	 * For example we know it will only once appear in the page next to the primary add button on articles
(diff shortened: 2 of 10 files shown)


components/jsx/follow-plus-instant-alerts/main.scss
```diff
@@ -32,3 +32,15 @@
 	background-position: 50% 41%;
 	margin-left: 12px;
 }
+
+.n-myft-follow-button--instant-alerts--open[aria-pressed=true]::after {
```
asugar13, line 36: **Question**: Out of curiosity, why do we need to target `[aria-pressed=true]` ?

components/jsx/preferences-modal/index.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * This preference modal is part of a test
+ * Therefore we have built the positioning function to work within the known parameters of that test
+ * For example we know it will only appear next to the primary add button on articles
+ * If this was to be used in other locations this function would need further improvements
+ */
+const positionModal = ({ eventTrigger, preferencesModal }) => {
```
asugar13, line 7: **Note:** Please note that in the designs, the popup modal covers all letters in the article title
designs:
Screenshot 2023-09-04 at 13 49 27

implementation:
Screenshot 2023-09-04 at 13 50 22

components/jsx/follow-plus-instant-alerts/main.scss
GlynnPhillips: That tells us that the button has been pressed and therefore the arrow icon should be in the "open" state

components/jsx/preferences-modal/index.js
GlynnPhillips: Thanks for picking that up - I must admit I just guessed at the gap and forgot to come back and check it against the designs

GlynnPhillips: Fixed

asugar13 (approved):

PR #595, opened 2023-09-01
Allow users from remove topic from myFT

GlynnPhillips:
This branches off from #594 so the only commits relevant to this change are the last 4

### Description

Allow users who see the follow + instant alerts experience to remove the topic from their myFT from within the preference modal.

### Whats included

1. Removing the topic from their myFT account
2. Closing the modal when the removal is successful
3. Disabling the button whilst the action is being processed
4. Error handling so the user knows when its failed, and don't close the modal

### Screen recordings

https://github.com/Financial-Times/n-myft-ui/assets/524573/fe044057-f3f6-4c2f-9384-3af917eaa1a3

### DIFF
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -6,8 +6,46 @@
  */
 const preferencesModal = document.querySelector('[data-component-id="myft-preferences-modal"]');
 
-const preferenceModalShowAndHide = () => {
+/**
+ * This preference modal is part of a test
+ * Therefore we have built the positioning function to work within the known parameters of that test
+ * For example we know it will only appear next to the primary add button on articles
+ * If this was to be used in other locations this function would need further improvements
+ */
+const positionModal = (eventTrigger) => {
+	const modalHorizontalCentering = (eventTrigger.offsetLeft + (eventTrigger.offsetWidth / 2)) - (preferencesModal.offsetWidth / 2);
+	const verticalPadding = 20;
+	const leftPadding = '5px';
+
+	preferencesModal.style.top = `${eventTrigger.offsetTop + eventTrigger.offsetHeight + verticalPadding}px`;
+	preferencesModal.style.left = `${modalHorizontalCentering}px`;
+
+	const modalPositionRelativeToScreen = preferencesModal.getBoundingClientRect();
+
+	if (modalPositionRelativeToScreen.left < 0) {
+		preferencesModal.style.left = leftPadding;
+	}
+
+	if (modalPositionRelativeToScreen.right > window.screen.width) {
+		const triggerRightPosition = eventTrigger.offsetLeft + eventTrigger.offsetWidth;
+		const modalShiftLeftPosition = triggerRightPosition - preferencesModal.offsetWidth;
+
+		preferencesModal.style.left = modalShiftLeftPosition > 0
+			? `${modalShiftLeftPosition}px`
+			: leftPadding;
+	}
+};
+
+const preferenceModalShowAndHide = (event) => {
+	if (!event.target) {
+		return;
+	}
+
 	preferencesModal.classList.toggle('n-myft-ui__preferences-modal--show');
+
+	if (preferencesModal.classList.contains('n-myft-ui__preferences-modal--show')) {
+		positionModal(event.target);
+	}
 };
 
 export default () => {
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -1,0 +1,19 @@
+/**
+ * This feature is part of a test
+ * Therefore we have built it to work within the known parameters of that test
+ * For example we know it will only once appear in the page next to the primary add button on articles
+ * If this was to be used in other locations it would need some additional work to avoid being singleton
+ */
+const preferencesModal = document.querySelector('[data-component-id="myft-preferences-modal"]');
+
+const preferenceModalShowAndHide = () => {
+	preferencesModal.classList.toggle('n-myft-ui__preferences-modal--show');
+};
+
+export default () => {
+	if (!preferencesModal) {
+		return;
+	}
+
+	document.addEventListener('myft.preference-modal.show-hide.toggle', preferenceModalShowAndHide);
+};
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -55,6 +55,8 @@ const preferenceModalShowAndHide = (event) => {
 };
 
 const removeTopic = async (event) => {
+	event.target.setAttribute('disabled', true);
+
 	try {
 		await myFtClient.remove('user', null, 'followed', 'concept', conceptId, { token: csrfToken });
 
@@ -62,6 +64,9 @@ const removeTopic = async (event) => {
 
 	} catch (error) {
 	}
+
+	event.target.removeAttribute('disabled');
+
 }
 
 export default () => {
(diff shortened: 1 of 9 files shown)


metart43:
@GlynnPhillips Sorry, I haven't tested your PR locally yet, but I wanted to ask a couple of questions. Modals (Dialogs) are very tricky to implement in terms of accessibility. I have never done that myself and most of the time relied on 3-rd party libraries. I glanced through the [w3 docs](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) and noticed there are rules around keyboard navigation and dialogues

> Keyboard Interaction
In the following description, the term "tabbable element" refers to any element with a tabindex value of zero or greater. Note that values greater than 0 are strongly discouraged.
 When a dialog opens, focus moves to an element inside the dialog. See notes below regarding initial focus placement.
    Tab:
        Moves focus to the next tabbable element inside the dialog.
        If focus is on the last tabbable element inside the dialog, moves focus to the first tabbable element inside the dialog.
    Shift + Tab:
        Moves focus to the previous tabbable element inside the dialog.
        If focus is on the first tabbable element inside the dialog, moves focus to the last tabbable element inside the dialog.
    Escape: Closes the dialog.

I obviously don't know if accessibility is a blocker in terms of this feature release.

components/jsx/preferences-modal/index.js
```diff
@@ -0,0 +1,92 @@
+import myFtClient from 'next-myft-client';
+import getToken from '../../../myft/ui/lib/get-csrf-token';
+
+const csrfToken = getToken();
```
metart43, line 4: what does `csrf` abbreviation stand for?

components/jsx/preferences-modal/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+	try {
+		await myFtClient.remove('user', null, 'followed', 'concept', conceptId, { token: csrfToken });
+
+		preferenceModalShowAndHide({ preferencesModal });
+
+	} catch (error) {
+		const parentNode = event.target.parentNode;
+		const errorElement = document.createElement('span');
+		errorElement.innerHTML = 'Sorry, we are unable to remove this topic. Please try again later or try from ft.com/myft';
```
asugar13, line 61: **Praise**: Good call on adding an error message. Did product/design approve it?

metart43 (commented):
The implementation looks good to me. Nice work on this. As I mentioned to you before modals are very ticky and annoying UI components to build from scratch. I just added 1 suggestion

components/jsx/preferences-modal/index.js
GlynnPhillips: Cross Site Request Forgery 

https://owasp.org/www-community/attacks/csrf

components/jsx/preferences-modal/index.js
GlynnPhillips: They haven't seen this yet, I intended to get their feedback after and iterate on it because they were off last week. Now that they are back maybe its worth involving them tomorrow

components/jsx/preferences-modal/index.js
```diff
@@ -0,0 +1,92 @@
...
+		errorElement.setAttribute('aria-live', 'polite');
+		errorElement.classList.add('n-myft-ui__preferences-modal-error');
+
+		parentNode.insertBefore(errorElement, event.target);
```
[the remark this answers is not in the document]
GlynnPhillips: Ahh yes I saw this earlier as well. I need to add an improvement to cover this :)

components/jsx/preferences-modal/preferences-modal.jsx
```diff
@@ -14,27 +14,27 @@ import React from 'react';
...
-export default function InstantAlertsPreferencesModal({ flags, currentPreferences }) {
+export default function InstantAlertsPreferencesModal({ conceptId, flags, currentPreferences, visible }) {
 	if (!flags.myFtApiWrite) {
 		return null;
 	}
 
 	const formattedCurrentPreferences = currentPreferences.join(', ')
 	return (
-		<div className="n-myft-ui__preferences-modal">
+		<div className={`n-myft-ui__preferences-modal ${visible ? 'n-myft-ui__preferences-modal--show' : ''}`} data-component-id="myft-preferences-modal" data-concept-id={conceptId}>
```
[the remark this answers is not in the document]
GlynnPhillips: Great suggestions. I will do some further research as I think some of this is useful but maybe not all of it. For example I want to better understand if theres any advantage to `aria-modal` vs `role=dialog`

components/jsx/preferences-modal/index.js
```diff
@@ -29,17 +40,38 @@ const positionModal = ({ eventTrigger, preferencesModal }) => {
...
-		positionModal({ eventTrigger: event.target, preferencesModal });
+		positionModal({ event, preferencesModal });
+	} else {
+		// Remove existing errors when hiding the modal
+		renderError('');
 	}
 };
 
+const renderError = ({ message, preferencesModal }) => {
+	const errorElement = preferencesModal.querySelector('[data-component-id="myft-preference-modal-error"]');
```
asugar13, line 54: Non blocking thought: maybe we can return early if `errorElement` has not been found

components/jsx/preferences-modal/index.js
```diff
@@ -29,17 +40,38 @@ const positionModal = ({ eventTrigger, preferencesModal }) => {
...
+const removeTopic = async ({ event, conceptId, preferencesModal }) => {
+	event.target.setAttribute('disabled', true);
+
+	try {
+		await myFtClient.remove('user', null, 'followed', 'concept', conceptId, { token: csrfToken });
+
+		preferenceModalShowAndHide({ preferencesModal });
+
+	} catch (error) {
+		renderError({ message: 'Sorry, we are unable to remove this topic. Please try again later or try from <a href="/myft">your myFT</a>', preferencesModal });
```
asugar13, line 68: **Thought**: Let's confirm this with product but I think just `myFT` without `your` would look better and on a single line.
Screenshot 2023-09-06 at 14 43 18

GlynnPhillips: Opps I made that change myself as well but forgot to push it

components/jsx/preferences-modal/index.js
GlynnPhillips: Yeah I am actually trying to think if this is the right thing to do or not (along with some other places where I have exited early)

It might be helpful to let it error/throw in some cases so we can better identify the errors instead of them becoming silent/unnoticed

GlynnPhillips: So lets discuss this later and come up with an approach

components/jsx/preferences-modal/index.js
```diff
@@ -29,17 +40,38 @@ const positionModal = ({ eventTrigger, preferencesModal }) => {
...
-	}
-
 	preferencesModal.classList.toggle('n-myft-ui__preferences-modal--show');
 
 	if (preferencesModal.classList.contains('n-myft-ui__preferences-modal--show')) {
-		positionModal({ eventTrigger: event.target, preferencesModal });
+		positionModal({ event, preferencesModal });
+	} else {
+		// Remove existing errors when hiding the modal
+		renderError('');
```
ytcleon, line 49: I think we need to specifc the `message` and `preferencesModal` here
```suggestion
		renderError({
			message: '',
			preferencesModal,
		});
```

components/jsx/preferences-modal/index.js
asugar13: Agreed. But where would we log these errors?

GlynnPhillips: Currently client side errors still go to Sentry (I think)

asugar13 (approved):
looks good!
note: I think you ran git pull origin main (without the --rebase) and that's why there's an extra commit

ytcleon (approved):

PR #599, opened 2023-09-07
PT-237 Show browser option in preferences modal

asugar13:
### Description:
Adds the browser option to the preferences modal when the user has browser notifications enabled for instant alerts.
This code is "heavily inspired" by the code in `next-myft-page` that ticks the checkbox for "Notifications on this device" :https://github.com/Financial-Times/next-myft-page/blob/main/client/components/alerts/js/push-notifications.js#L53

this PR extends #597 
JIRA: https://financialtimes.atlassian.net/jira/software/c/projects/PT/boards/1655?modal=detail&selectedIssue=PT-237

https://github.com/Financial-Times/n-myft-ui/assets/10453619/925fdbd7-1f75-4323-b72b-e79fece5268e

### DIFF
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -74,7 +74,7 @@ const removeTopic = async ({ event, conceptId, preferencesModal }) => {
 	event.target.removeAttribute('disabled');
 };
 
-const getAlertsPreferences = ({ event, preferencesModal }) => {
+const getAlertsPreferences = async ({ event, preferencesModal }) => {
 	const preferencesList = preferencesModal.querySelector('[data-component-id="myft-preferences-modal-list"]');
 
 	if (!preferencesList) {
@@ -91,11 +91,23 @@ const getAlertsPreferences = ({ event, preferencesModal }) => {
 		}
 	});
 
-	const alertsEnabledText = preferencesList.innerHTML + addedTextBuffer.join(',') + '.';
-	const alertsDisabledText = 'You have disabled all instant alerts';
+	try {
+		// We need the service worker registration to check for a subscription
+		const serviceWorkerRegistration = await navigator.serviceWorker.ready;
+		const subscription = await serviceWorkerRegistration.pushManager.getSubscription();
+		if (subscription) {
+			addedTextBuffer.push('browser');
+		}
 
+	} catch (error) {
+		// eslint-disable-next-line no-console
+		console.warn('There was an error fetching the browser notification preferences', error);
+	}
+	const alertsEnabledText = `Your delivery channels: ${addedTextBuffer.join(', ')}.`;
+	const alertsDisabledText = 'You have previously disabled all delivery channels';
 	preferencesList.innerHTML = addedTextBuffer.length > 0 ? alertsEnabledText : alertsDisabledText;
-}
+
+};
 
 const setCheckboxForAlertConcept = ({ event, preferencesModal }) => {
 	const conceptId = preferencesModal.dataset.conceptId;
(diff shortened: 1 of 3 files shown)


components/jsx/preferences-modal/index.js
```diff
@@ -91,11 +91,26 @@ const getAlertsPreferences = ({ event, preferencesModal }) => {
 		}
 	});
 
-	const alertsEnabledText = preferencesList.innerHTML + addedTextBuffer.join(',') + '.';
-	const alertsDisabledText = 'You have disabled all instant alerts';
+	const alertsEnabledText = 'Your delivery channels: ' + addedTextBuffer.join(', ') + '.';
+	const alertsDisabledText = 'You have previously disabled all delivery channels';
 
 	preferencesList.innerHTML = addedTextBuffer.length > 0 ? alertsEnabledText : alertsDisabledText;
```
GlynnPhillips, line 97: **Comment/suggestion:** Could all of the text rendering be moved after the awaited push notification code so we are only rendering once?

Something like this

```js
try {
	// We need the service worker registration to check for a subscription
	const serviceWorkerRegistration = await navigator.serviceWorker.ready;
	const subscription = await serviceWorkerRegistration.pushManager.getSubscription();
	if (!subscription) {
		// We aren't subscribed to push, so return early
		return;
	}
	addedTextBuffer.push('browser');
} catch (error) {
	// eslint-disable-next-line no-console
	console.warn('There was an error fetching the browser notification preferences', error);
}

const alertsEnabledText = `Your delivery channels: ${addedTextBuffer.join(', ')}.`;
const alertsDisabledText = 'You have previously disabled all delivery channels';

preferencesList.innerHTML = addedTextBuffer.length > 0 ? alertsEnabledText : alertsDisabledText;
```

components/jsx/preferences-modal/index.js
```diff
@@ -91,11 +91,26 @@ const getAlertsPreferences = ({ event, preferencesModal }) => {
 		}
 	});
 
-	const alertsEnabledText = preferencesList.innerHTML + addedTextBuffer.join(',') + '.';
-	const alertsDisabledText = 'You have disabled all instant alerts';
+	const alertsEnabledText = 'Your delivery channels: ' + addedTextBuffer.join(', ') + '.';
```
GlynnPhillips, line 94: **Change request:** I think a template literal would be slightly nicer here

```suggestion
	const alertsEnabledText = `Your delivery channels: ${addedTextBuffer.join(', ')}.`;
```

GlynnPhillips (commented):
Some minor comments but looking good once addressed

components/jsx/preferences-modal/index.js
```diff
@@ -91,11 +91,26 @@ const getAlertsPreferences = ({ event, preferencesModal }) => {
...
+	const alertsDisabledText = 'You have previously disabled all delivery channels';
 
 	preferencesList.innerHTML = addedTextBuffer.length > 0 ? alertsEnabledText : alertsDisabledText;
-}
+
+	try {
+		// We need the service worker registration to check for a subscription
+		const serviceWorkerRegistration = await navigator.serviceWorker.ready;
+		const subscription = await serviceWorkerRegistration.pushManager.getSubscription();
```
ytcleon, line 102: *Change request*: I think we need to consider that the service worker is sometimes unavailable (e.g. the browser is in incognito mode)
```suggestion
		const serviceWorkerRegistration = await navigator.serviceWorker?.ready;
		const subscription = await serviceWorkerRegistration?.pushManager.getSubscription();
```

asugar13: That's a good point. However, in this case the error will be caught on line 102 and the browser option won't be added to the message. Also, it seems like the linter is showing warnings for optional chaining `?.`.
Let me know your thoughts.

ytcleon: That's a good point. Since the error is caught properly, I think it is fine to keep it as is .

PR #601, opened 2023-09-08
Instant alerts on off

GlynnPhillips:
Allow users to turn instant alerts on/off from within the preference modal component

### Whats included

1. Make the request to turn on/off instant alerts
2. Error handling for the request
3. Switching the state back when erroring
4. Switching the visual status of the alert icon in the follow button

### Jira

https://financialtimes.atlassian.net/browse/PT-236

### DIFF
--- a/components/jsx/follow-plus-instant-alerts/index.js
+++ b/components/jsx/follow-plus-instant-alerts/index.js
@@ -1,11 +1,34 @@
 const setBellForAlertConcept = ({ event, followPlusInstantAlerts }) => {
-	const conceptId = followPlusInstantAlerts.dataset.conceptId;
-	// search through all the concepts that the user has followed and check whether
-	// 1. the concept which this instant alert modal controls is within them, AND;
-	// 2. the said concept has instant alert enabled
-	// if so, set the bell on in the button
-	const currentConcept = event.detail.items.find(item => item && item.uuid === conceptId);
-	if (currentConcept && currentConcept._rel && currentConcept._rel.instant) {
+
+	if (!event) {
+		return;
+	}
+
+	/**
+	 * There are two types of event we receive
+	 * update - happens when the user changes their follows/instant alert
+	 * load - happens when the page loads and contains all follow data
+	 * The payload of those events are different so we need to handle them differently
+	 */
+	const isUpdateEvent = event.type === 'myft.user.followed.concept.update';
+	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
+
+	// Search through the event payload to see if any data relates to this modal
+	const currentConcept = isUpdateEvent
+		? event.detail.results.find(item => item && item.subject && item.subject.properties.uuid === modalConceptId)
+		: event.detail.items.find(item => item && item.uuid === modalConceptId);
+
+	if (!currentConcept) {
+		return;
+	}
+
+	// Does the event indicate that instant alerts are turned on
+	const isTurningAlertsOn = isUpdateEvent
+		? currentConcept.rel && currentConcept.rel.properties && currentConcept.rel.properties.instant
+		: currentConcept._rel && currentConcept._rel.instant;
+
+	// Update the button icon to reflect the instant alert preference
+	if (isTurningAlertsOn) {
 		followPlusInstantAlerts.classList.add('n-myft-follow-button--instant-alerts--on');
 	} else {
 		followPlusInstantAlerts.classList.remove('n-myft-follow-button--instant-alerts--on');
@@ -36,4 +59,5 @@ export default () => {
 	followPlusInstantAlerts.addEventListener('click', () => sendModalToggleEvent({followPlusInstantAlerts}));
 
 	document.body.addEventListener('myft.user.followed.concept.load', (event) => setBellForAlertConcept({event, followPlusInstantAlerts}));
+	document.body.addEventListener('myft.user.followed.concept.update', (event) => setBellForAlertConcept({event, followPlusInstantAlerts}));
 };
--- a/components/jsx/follow-plus-instant-alerts/index.js
+++ b/components/jsx/follow-plus-instant-alerts/index.js
@@ -1,39 +1,52 @@
-const setBellForAlertConcept = ({ event, followPlusInstantAlerts }) => {
-
-	if (!event) {
-		return;
-	}
-
-	/**
-	 * There are two types of event we receive
-	 * update - happens when the user changes their follows/instant alert
-	 * load - happens when the page loads and contains all follow data
-	 * The payload of those events are different so we need to handle them differently
-	 */
-	const isUpdateEvent = event.type === 'myft.user.followed.concept.update';
-	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
-
-	// Search through the event payload to see if any data relates to this modal
-	const currentConcept = isUpdateEvent
-		? event.detail.results.find(item => item && item.subject && item.subject.properties.uuid === modalConceptId)
-		: event.detail.items.find(item => item && item.uuid === modalConceptId);
-
-	if (!currentConcept) {
-		return;
-	}
-
-	// Does the event indicate that instant alerts are turned on
-	const isTurningAlertsOn = isUpdateEvent
-		? currentConcept.rel && currentConcept.rel.properties && currentConcept.rel.properties.instant
-		: currentConcept._rel && currentConcept._rel.instant;
-
+const toggleInstantAlertsClass = ({ instantAlertsOn,followPlusInstantAlerts }) => {
 	// Update the button icon to reflect the instant alert preference
-	if (isTurningAlertsOn) {
+	if (instantAlertsOn) {
 		followPlusInstantAlerts.classList.add('n-myft-follow-button--instant-alerts--on');
 	} else {
 		followPlusInstantAlerts.classList.remove('n-myft-follow-button--instant-alerts--on');
 	}
 };
+
+const instantAlertsIconLoad = ({ event, followPlusInstantAlerts }) => {
+	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
+
+	if (!event || !modalConceptId) {
+		return;
+	}
+
+	const currentConcept = event.detail.items
+		.find(item => item && item.uuid === modalConceptId);
+
+	if (!currentConcept) {
+		return;
+	}
+
+
+	const instantAlertsOn = Boolean(currentConcept && currentConcept._rel && currentConcept._rel.instant);
+	toggleInstantAlertsClass({instantAlertsOn, followPlusInstantAlerts });
+};
+
+const instantAlertsIconUpdate = ({ event, followPlusInstantAlerts }) => {
+	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
+
+	if (!event || !modalConceptId) {
+		return;
+	}
+
+	const currentConcept = event.detail.results
+		.find(item => item && item.subject && item.subject.properties && item.subject.properties.uuid === modalConceptId);
+
+	if (!currentConcept) {
+		return;
+	}
+
+
+	const instantAlertsOn = Boolean(currentConcept && currentConcept.rel && currentConcept.rel.properties && currentConcept.rel.properties.instant);
+	toggleInstantAlertsClass({instantAlertsOn, followPlusInstantAlerts });
+};
+
+
+
 
 const sendModalToggleEvent = ({ followPlusInstantAlerts }) => {
 	const preferenceModalToggleEvent = new CustomEvent('myft.preference-modal.show-hide.toggle', { bubbles: true });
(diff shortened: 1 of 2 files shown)


components/jsx/follow-plus-instant-alerts/index.js
```diff
@@ -1,11 +1,34 @@
 const setBellForAlertConcept = ({ event, followPlusInstantAlerts }) => {
-	const conceptId = followPlusInstantAlerts.dataset.conceptId;
-	// search through all the concepts that the user has followed and check whether
-	// 1. the concept which this instant alert modal controls is within them, AND;
-	// 2. the said concept has instant alert enabled
-	// if so, set the bell on in the button
-	const currentConcept = event.detail.items.find(item => item && item.uuid === conceptId);
-	if (currentConcept && currentConcept._rel && currentConcept._rel.instant) {
+
+	if (!event) {
```
asugar13, line 3: **non-blocking nitpick:** I know we've talked about this before but every time I see a "return early" statement I think "it'd be nice to do this in one line"

components/jsx/follow-plus-instant-alerts/index.js
```diff
@@ -1,11 +1,34 @@
...
+	 * update - happens when the user changes their follows/instant alert
+	 * load - happens when the page loads and contains all follow data
+	 * The payload of those events are different so we need to handle them differently
+	 */
+	const isUpdateEvent = event.type === 'myft.user.followed.concept.update';
+	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
+
+	// Search through the event payload to see if any data relates to this modal
+	const currentConcept = isUpdateEvent
+		? event.detail.results.find(item => item && item.subject && item.subject.properties.uuid === modalConceptId)
```
asugar13, line 18: **Note:** It'd be great if we could use optional chaining. But since we can't, I think this is missing  checking that .properties is defined and not null
item && item.subject && item.subject.properties && item.subject.properties.uuid

components/jsx/follow-plus-instant-alerts/index.js
```diff
@@ -1,11 +1,34 @@
...
+	}
+
+	/**
+	 * There are two types of event we receive
+	 * update - happens when the user changes their follows/instant alert
+	 * load - happens when the page loads and contains all follow data
+	 * The payload of those events are different so we need to handle them differently
+	 */
+	const isUpdateEvent = event.type === 'myft.user.followed.concept.update';
+	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
```
asugar13, line 14: **Thought:** The conceptId is retrieved from the default function in the preferences-modal

components/jsx/follow-plus-instant-alerts/index.js
```diff
@@ -1,11 +1,34 @@
...
+
+	// Search through the event payload to see if any data relates to this modal
+	const currentConcept = isUpdateEvent
+		? event.detail.results.find(item => item && item.subject && item.subject.properties.uuid === modalConceptId)
+		: event.detail.items.find(item => item && item.uuid === modalConceptId);
```
asugar13, line 19: **Thought**: Didn't find it too obvious with the logical statement using `isUpdateEvent`. Perhaps we can wrap these in an if/else? Or what about have 2 functions for the 2 event listeners? The functions could be called `setBellForAlertConcept` and `updateBellForAlertConcept`

components/jsx/follow-plus-instant-alerts/index.js
```diff
@@ -1,11 +1,34 @@
...
+
+	if (!currentConcept) {
+		return;
+	}
+
+	// Does the event indicate that instant alerts are turned on
+	const isTurningAlertsOn = isUpdateEvent
+		? currentConcept.rel && currentConcept.rel.properties && currentConcept.rel.properties.instant
```
asugar13, line 27: similar note as above: https://github.com/Financial-Times/n-myft-ui/pull/601/files#r1319890846

asugar13 (commented):
looks good just left a few thoughts/suggestions

components/jsx/follow-plus-instant-alerts/index.js
ytcleon: I agree with @asugar13 idea. With two separate functions it is easier to understand what value we want to extract from `currentConcept` for different events.

components/jsx/follow-plus-instant-alerts/index.js
ytcleon: **Question**: I am a bit confused on the naming of the field that contains the instant alert value. For the update event, the field is called `rel`, while for the load event, the field is called `_rel`. Is it correct?

components/jsx/follow-plus-instant-alerts/index.js
GlynnPhillips: When I see it on one line I think "It would be nice to use braces" 😂 differences of opinion and I guess this is what linting is there for so individuals don't need to worry. However I do find multiple lines / braces easier to read 🤷 I wouldn't block a PR based on this though - I would leave that down to linting

components/jsx/follow-plus-instant-alerts/index.js
GlynnPhillips: Good spot 👍

components/jsx/follow-plus-instant-alerts/index.js
GlynnPhillips: >  why is the variable name modalConceptId instead of just conceptId

Seeing as we are dealing with multiple concept Ids at this stage I just thought it was more descriptive as to which concept id it was.

> The conceptId is retrieved from the default function in the preferences-modal

Yes this was already in place here I just renamed it. I think we need to revisit how we access these probs etc so we don't need to pass them around as much but we can maybe do this as a follow up

components/jsx/follow-plus-instant-alerts/index.js
GlynnPhillips: > the field is called rel, while for the load event, the field is called _rel. Is it correct?


Yes unfortunately this is one of the differences of the event types which means we need to treat them differently :(

components/jsx/preferences-modal/index.js
```diff
@@ -112,6 +112,37 @@ const setCheckboxForAlertConcept = ({ event, preferencesModal }) => {
 	}
 };
 
+const toggleInstantAlertsPreference = async ({ event, conceptId, preferencesModal }) => {
+	const instantAlertsToggle = event.target;
+
+	instantAlertsToggle.setAttribute('disabled', true);
```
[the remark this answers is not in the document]
GlynnPhillips: Can you explain which indent is missing? I am not sure I am seeing it

components/jsx/follow-plus-instant-alerts/index.js
GlynnPhillips: I am just mocking up an example response for this

GlynnPhillips: Ok, this is not perfect but I thought I would see what you think of this concept before I tidy it up. What about reformatting the event payloads first so both load and update events have the same structure. Then the code afterwards gets simplier because it can handle them the same each time.


```js
const formatEventPayload = (event) => {
	/**
	 * There are two types of event we receive
	 * update - happens when the user changes their follows/instant alert
	 * load - happens when the page loads and contains all follow data
	 * The payload of those events are different so we to reform them to match
	 */

	const newStructure = ({ conceptId, instantAlertsOn }) => {
		if (!conceptId) {
			return false;
		}

		return {
			uuid: conceptId,
			instantAlertsOn
		}
	};

	if (event.type === 'myft.user.followed.concept.update') {
		return event.detail.results
			.map(item => {
				const conceptId = item && item.subject && item.subject.properties && item.subject.properties.uuid;
				const instantAlertsOn = Boolean(item && item.subject && item.subject.properties && item.subject.properties.instant);

				return newStructure({ conceptId, instantAlertsOn });
			});
	} else if (event.type === 'myft.user.followed.concept.load') {
		return event.detail.items
			.map(item => {
				const conceptId = item && item.uuid;
				const instantAlertsOn = Boolean(item && item._rel && item._rel.instant);

				return newStructure({ conceptId, instantAlertsOn });
			});
	}
};

const getMatchingEventConcept = ({ eventPayload, modalConceptId }) => {
	if (!eventPayload.length) {
		return;
	}

	return eventPayload.find(item => item.uuid === modalConceptId);
};


const setBellForAlertConcept = ({ event, followPlusInstantAlerts }) => {

	if (!event) {
		return;
	}

	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
	const reformattedEventPayload = formatEventPayload(event);

	const currentConcept = getMatchingEventConcept({ eventPayload: reformattedEventPayload, modalConceptId });

	// Update the button icon to reflect the instant alert preference
	if (currentConcept && currentConcept.instantAlertsOn) {
		followPlusInstantAlerts.classList.add('n-myft-follow-button--instant-alerts--on');
	} else {
		followPlusInstantAlerts.classList.remove('n-myft-follow-button--instant-alerts--on');
	}
};
```

components/jsx/follow-plus-instant-alerts/index.js
asugar13: Since we're returning early a lot and not catching/logging errors that much now, I think it may make sense to move this to default https://github.com/Financial-Times/n-myft-ui/blob/main/components/jsx/follow-plus-instant-alerts/index.js#L32 and return early for consistency. 
Don't have a strong opinion
We can create a ticket to revisit this and catching/logging errros.

components/jsx/follow-plus-instant-alerts/index.js
asugar13: I think the code would be more readable if we simply had 2 functions for this:  -one for setting the bell icon on load 
-one for updating the bell icon 

I think it's better to have 2 functions. There are 2 different myFT events. One of them retrieves all concepts when the page is loaded, and the other one retrieves one item when it's updated.
We wouldn't need for intermediate functions such as `newStructure` (I also think `formatConcept` for the function name would be a bit clearer)

I understand the code now so it makes sense. But I was a bit thrown off at first glance. I think supporting multiple events in one function is an over-optimisation for now.

components/jsx/preferences-modal/index.js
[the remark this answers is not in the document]
asugar13: please ignore, I guess I was looking at this on a small window size 😅

components/jsx/follow-plus-instant-alerts/index.js
GlynnPhillips: I haven't tested this fully yet but would this be better

In the default function

```js
document.body.addEventListener('myft.user.followed.concept.load', (event) => instantAlertsIconLoad({event, followPlusInstantAlerts}));
document.body.addEventListener('myft.user.followed.concept.update', (event) => instantAlertsIconUpdate({event, followPlusInstantAlerts}));
```

and then

```js
const toggleInstantAlertsClass = ({ instantAlertsOn,followPlusInstantAlerts }) => {
	// Update the button icon to reflect the instant alert preference
	if (instantAlertsOn) {
		followPlusInstantAlerts.classList.add('n-myft-follow-button--instant-alerts--on');
	} else {
		followPlusInstantAlerts.classList.remove('n-myft-follow-button--instant-alerts--on');
	}
}

const instantAlertsIconLoad = ({ event, followPlusInstantAlerts }) => {

	if (!event) {
		return;
	}

	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
	const currentConept = event.detail.items
		.find(item => item && item.uuid === modalConceptId);

	if (!currentConept) {
		return;
	}


	const instantAlertsOn = Boolean(currentConept && currentConept._rel && currentConept._rel.instant);
	toggleInstantAlertsClass({instantAlertsOn, followPlusInstantAlerts })
};

const instantAlertsIconUpdate = ({ event, followPlusInstantAlerts }) => {

	if (!event) {
		return;
	}

	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
	const currentConept = event.detail.items
		.find(item => item && item.subject && item.subject.properties && item.subject.properties.uuid === modalConceptId);

	if (!currentConept) {
		return;
	}


	const instantAlertsOn = Boolean(currentConept && currentConept.subject && currentConept.subject.properties && currentConept.subject.properties.instant);
	toggleInstantAlertsClass({instantAlertsOn, followPlusInstantAlerts })
};

```

asugar13: This looks good for the most part!

asugar13: feel free to return early if `currentConept.uuid` is undefined or if `modalConceptId` is undefined (before iterating over .`event.detail.items`)

ytcleon: I like this version, which is more readable and easier to debug with the stack trace.

components/jsx/follow-plus-instant-alerts/index.js
```diff
@@ -1,17 +1,53 @@
...
+	if (!currentConcept) {
+		return;
+	}
+
+
+	const instantAlertsOn = Boolean(currentConcept && currentConcept.rel && currentConcept.rel.properties && currentConcept.rel.properties.instant);
+	toggleInstantAlertsClass({instantAlertsOn, followPlusInstantAlerts });
 };
 
+
```
asugar13, line 48: nitpick: it's a big jump line-wise!

asugar13 (approved):

PR #605, opened 2023-09-14
PT-278 Add JSX transpiler

asugar13:
### Description:
When publishing n-myft-ui and installing from next-article, we've realised that the JSX components were not being transpiled into babel 6.
This PR creates a custom babel environment to be used in the make transpile-jsx step which has been added to the circleci config.yml file. The custom babel environment runs Babel 6.

[JIRA](https://financialtimes.atlassian.net/jira/software/c/projects/PT/boards/1655?modal=detail&selectedIssue=PT-278)


.gitignore
```diff
@@ -9,4 +9,5 @@
 public/
 pa11y_screenCapture/
 .idea/
-.vscode/
\ No newline at end of file
+.vscode/
+dist/
```
asugar13, line 13: **comment:** We may need to add this to .editorconfig as well to prevent errors from lintspace.

PR #609, opened 2023-09-18
PT-286 Fix return early statements

asugar13:
### Description

Following up on this incident: https://financialtimes.slack.com/archives/C05T4CC8MTK
We merged this [PR](https://github.com/Financial-Times/next-article/pull/5208) which caused the following error:  Uncaught (in promise) TypeError: Cannot read properties of null (reading 'dataset')

What happened:
The clientside JS was [always running in next-article](https://github.com/Financial-Times/next-article/pull/5208/files#diff-24d413c6815f76e41fc78f7656ef33cb2d524937ec0abbb1e4a55090b258e3d2R247). This didn't require the [myFTInstantAlertsOnboarding](https://toggler.ft.com/#myFTInstantAlertsOnboarding) flag to be on. Which caused that error to trigger when the JS was ran and the preferencesModal was not rendered

[JIRA](https://financialtimes.atlassian.net/browse/PT-286)

### DIFF
--- a/components/jsx/follow-plus-instant-alerts/index.js
+++ b/components/jsx/follow-plus-instant-alerts/index.js
@@ -8,7 +8,7 @@ const toggleInstantAlertsClass = ({ instantAlertsOn,followPlusInstantAlerts }) =
 };
 
 const instantAlertsIconLoad = ({ event, followPlusInstantAlerts }) => {
-	const modalConceptId = followPlusInstantAlerts.dataset && followPlusInstantAlerts.dataset.conceptId;
+	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
 
 	if (!event || !modalConceptId) {
 		return;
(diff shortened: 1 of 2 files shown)


components/jsx/follow-plus-instant-alerts/index.js
```diff
@@ -8,7 +8,7 @@ const toggleInstantAlertsClass = ({ instantAlertsOn,followPlusInstantAlerts }) =
 };
 
 const instantAlertsIconLoad = ({ event, followPlusInstantAlerts }) => {
-	const modalConceptId = followPlusInstantAlerts.dataset.conceptId;
+	const modalConceptId = followPlusInstantAlerts.dataset && followPlusInstantAlerts.dataset.conceptId;
```
GlynnPhillips, line 11: **Question:** Whats the reason for checking that `dataset` exists here?

asugar13: Mostly me being a bit paranoid after that incident. If the element's passed then it should have a `.dataset` attribute though (unless a different element is passed to the function)

Should I remove the checking?

GlynnPhillips (approved):

PR #610, opened 2023-09-18
PT-285 Render alerts preferences available before waiting for browser alerts

asugar13:
### Description
See this piece of documentation: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/ready
As you can see `.ready` for the ServiceWorkerContainer "waits indefinitely until the [ServiceWorkerRegistration](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration) associated with the current page has an [active](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/active) worker"

This meant that when no active worker was found, nothing was being rendered in the preferences modal to show the user's alert preferences.

[JIRA](https://financialtimes.atlassian.net/browse/PT-285)

Before:
Screenshot 2023-09-18 at 13 09 50

After:
Screenshot 2023-09-18 at 10 17 44

### DIFF
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -74,6 +74,12 @@ const removeTopic = async ({ event, conceptId, preferencesModal }) => {
 	event.target.removeAttribute('disabled');
 };
 
+const getAlertsPreferenceText = (addedTextBuffer) => {
+	const alertsEnabledText = `Your delivery channels: ${addedTextBuffer.join(', ')}.`;
+	const alertsDisabledText = 'You have previously disabled all delivery channels';
+	return Array.isArray(addedTextBuffer) && addedTextBuffer.length > 0 ? alertsEnabledText : alertsDisabledText;
+};
+
 const getAlertsPreferences = async ({ event, preferencesModal }) => {
 	const preferencesList = preferencesModal.querySelector('[data-component-id="myft-preferences-modal-list"]');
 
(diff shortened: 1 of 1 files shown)


components/jsx/preferences-modal/index.js
```diff
@@ -91,20 +91,23 @@ const getAlertsPreferences = async ({ event, preferencesModal }) => {
 		}
 	});
 
+	let alertsEnabledText = `Your delivery channels: ${addedTextBuffer.join(', ')}.`;
+	const alertsDisabledText = 'You have previously disabled all delivery channels';
```
asugar13, line 95: **Note for reviewer**: I am aware it's not great to paint `preferencesList.innerHTML` twice but for now it's the only I've come up to render something without having to wait for the serviceWorker calls to resolve

ytcleon: **Suggestion**: The two scripts to generate the text in the preference modal are exactly the same. We can put them in a small helper function.
```
function getAlertsText(addedTextBuffer) {
  const alertsEnabledText = `Your delivery channels: ${addedTextBuffer.join(', ')}.`;
  const alertsDisabledText = 'You have previously disabled all delivery channels';
  // it is fine not to check whether addedTextBuffer is array within getAlertsPreferences. I added it just as a precaution.
  return Array.isArray(addedTextBuffer) && addedTextBuffer.length > 0 ? alertsEnabledText : alertsDisabledText;
}
// ...
preferencesList.innerHTML = getAlertsText(addedTextBuffer);
```

ytcleon (approved):

PR #613, opened 2023-09-25
CON-2746 Update to node 18

camillecroci:
# Description

PR to achieve upgrade to Node v18. This is part of [migrating to Node v.18 Initiative](https://financialtimes.atlassian.net/browse/CON-2706).

In order to achieve the migration, it has been used the following [migration script](https://github.com/Financial-Times/platform-scripts/tree/main/upgrade-node-18).

# Petit plus

There was some updated dependencies and the famous `--openssl-legacy-provider` added because webpack (and webpack in karma) have some issues with node 18.  
Because node 16 is meant to be supported (why ? I don't know and disagree with it but its in the migration guide), our CI runs tests on node 16 and so there is a bit of code in the makefile that checks the node version to add the flag only when running node 18. The flag thingy was failing with node 16 

# Test

To test that it would not break anything, I plugged that branch to **next-stream-page** (which had a succesful update to node 18). 
I checked if the `init` ([code](https://github.com/Financial-Times/n-myft-ui/blob/main/myft/ui/index.js#L6C1-L6C1)) was launched properly by adding a console.log.  

Screenshot 2023-09-25 at 17 58 44

I tested the components by adding a little thing in the form around the button. 
Screenshot 2023-09-25 at 17 59 17

I could see them both. 

I toggled a few stream page and added/removed a topic from a stream page (to my ft) and did not see anything weird :) 
I also ran the tests on **next-stream-page**


Makefile
```diff
@@ -4,6 +4,12 @@ node_modules/@financial-times/n-gage/index.mk:
 
 -include node_modules/@financial-times/n-gage/index.mk
 
+NODE_VERSION := $(shell node --version)
```
juancancela, line 7: Nice that this is working!

juancancela (approved):
lgtm!

Makefile
camillecroci: yes ! thanks again for that !

PR #614, opened 2023-09-25
PT-292 Hide modal when document is clicked (and it's not the preferences modal)

asugar13:
### Description
Up until now, the only to close the preference modal was to click on the addToMyFT button. This PR makes changes so that the modal is also closed when the user clicks elsewhere on the screen (but not on the preference modal itself)

[JIRA](https://financialtimes.atlassian.net/jira/software/c/projects/PT/boards/1655?modal=detail&selectedIssue=PT-292)

before:
https://github.com/Financial-Times/n-myft-ui/assets/10453619/a84209f1-b111-49aa-8499-d26be4da1219

after:
https://github.com/Financial-Times/n-myft-ui/assets/10453619/0d7b5a15-b85e-4cc5-8046-7b6736bb1010

### DIFF
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -75,6 +75,12 @@ const tracking = (conceptId) => {
 	document.body.dispatchEvent(trackingEvent);
 };
 
+const preferenceModalHide = ({ event, preferencesModal }) => {
+	if (!preferencesModal.contains(event.detail.targetElement)) {
+		preferencesModal.classList.remove('n-myft-ui__preferences-modal--show');
+	}
+};
+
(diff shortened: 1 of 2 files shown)


components/jsx/preferences-modal/index.js
```diff
@@ -75,6 +75,12 @@ const tracking = (conceptId) => {
 	document.body.dispatchEvent(trackingEvent);
 };
 
+const preferenceModalHide = ({ event, preferencesModal }) => {
+	if (!preferencesModal.contains(event.detail.targetElement)) {
```
ytcleon, line 79: **Praise**: Nice one on forwarding the click event target to the preference modal to do the checking, instead of checking the target inside the follow button and create coupling.

ytcleon (approved):

PR #620, opened 2023-10-16
Instant alert error message styling

ytcleon:
## Description

This PR is to add styling to the error message box within the preference modal of the instant alert. It adds the following styling changes:

- Error message box
  - top margin: 16px
  - padding: 8px for top/bottom, 12px for left/right
  - text color: #b30000 (color from oMessage)
  - background color: #fad9ce (color from oMessage)
  - text align: left

- Remove from myFT button when disabled:
  - opacity: 0.5

In addition, when the text inside the error message box changes, there is a CSS transition to emphasise the change.

## Screen capture

https://github.com/Financial-Times/n-myft-ui/assets/28527037/ad575f55-c050-44bc-84cb-d9b33b6a49f1

## Ticket:

[JIRA](https://financialtimes.atlassian.net/browse/ACC-2587)

### DIFF
--- a/components/jsx/preferences-modal/index.js
+++ b/components/jsx/preferences-modal/index.js
@@ -9,6 +9,13 @@ const renderError = ({ message, preferencesModal }) => {
 	errorElement.innerHTML = message;
 };
 
+const removeError = ({ preferencesModal }) => {
+	renderError({
+		message: "",
+		preferencesModal,
+	});
+}
+
 /**
  * This preference modal is part of a test
  * Therefore we have built the positioning function to work within the known parameters of that test
@@ -78,6 +85,7 @@ const tracking = (conceptId) => {
 const preferenceModalHide = ({ event, preferencesModal }) => {
 	if (!preferencesModal.contains(event.detail.targetElement)) {
 		preferencesModal.classList.remove('n-myft-ui__preferences-modal--show');
+		removeError({ preferencesModal });
 	}
 };
 
@@ -91,23 +99,25 @@ const preferenceModalShowAndHide = ({ event, preferencesModal, conceptId }) => {
 
 	} else {
 		// Remove existing errors when hiding the modal
-		renderError({
-			message: '',
-			preferencesModal,
-		});
+		removeError({ preferencesModal });
 	}
 };
 
 const removeTopic = async ({ event, conceptId, preferencesModal }) => {
 	event.target.setAttribute('disabled', true);
 
+	// remove error message before attempt to remove topic
+	removeError({ preferencesModal });
+
 	try {
 		await myFtClient.remove('user', null, 'followed', 'concept', conceptId, { token: csrfToken });
 		preferenceModalShowAndHide({ preferencesModal });
 	} catch (error) {
-		renderError({ message: 'Sorry, we are unable to remove this topic. Please try again later or try from <a href="/myft">myFT</a>', preferencesModal });
+		renderError({
+			message: 'Sorry, we are unable to remove this topic. Please try again later or try from <a href="/myft">myFT</a>',
+			preferencesModal,
+		});
 	}
-
 	event.target.removeAttribute('disabled');
 };
 
@@ -175,6 +185,9 @@ const toggleInstantAlertsPreference = async ({ event, conceptId, preferencesModa
 
 	instantAlertsCheckbox.setAttribute('disabled', true);
 
+	// remove error message before attempt to toggle instant alert
+	removeError({ preferencesModal });
+
 	const data = {
 		token: csrfToken
 	};
@@ -194,7 +207,7 @@ const toggleInstantAlertsPreference = async ({ event, conceptId, preferencesModa
 	} catch (error) {
 		renderError({
 			message: 'Sorry, we are unable to change your instant alert preference. Please try again later or try from <a href="/myft">myFT</a>',
-			preferencesModal
+			preferencesModal,
 		});
 
 		instantAlertsCheckbox.checked = instantAlertsCheckbox.checked
--- a/components/jsx/preferences-modal/main.scss
+++ b/components/jsx/preferences-modal/main.scss
@@ -47,14 +47,39 @@
 	color: oColorsByName('claret-60');
 }
 
+.n-myft-ui__preferences-modal__remove-button:disabled {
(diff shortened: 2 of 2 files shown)


components/jsx/preferences-modal/index.js
```diff
@@ -7,6 +7,7 @@ const renderError = ({ message, preferencesModal }) => {
 	const errorElement = preferencesModal.querySelector('[data-component-id="myft-preference-modal-error"]');
 
 	errorElement.innerHTML = message;
+	errorElement.style.display = 'block';
```
GlynnPhillips, line 10: **Question:** Why is this needed in the JS vs just doing it in the CSS?

components/jsx/preferences-modal/main.scss
```diff
@@ -38,6 +38,7 @@
 	background-color: oColorsByName('white-80');
 	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
 	color: oColorsByName('black-80');
+	opacity: .5;
```
GlynnPhillips, line 41: **Question:** Why has this been added to the default state of the button?

components/jsx/preferences-modal/main.scss
```diff
@@ -55,6 +56,10 @@
 .n-myft-ui__preferences-modal-error {
 	display: block;
 	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
-	color: oColorsByName('claret');
-	text-align: center;
+	color: #b30000;
```
GlynnPhillips, line 59: **Change request:** I think we should be able to use oColor here

components/jsx/preferences-modal/main.scss
```diff
@@ -55,6 +56,10 @@
...
+	background-color: #fad9ce;
```
GlynnPhillips, line 60: **Change request:** Again I think we should be able to use oColor here

components/jsx/preferences-modal/main.scss
ytcleon: I cannot find an exact match of the two colors in the o-colors palette. The closest I can find are wheat(#f2dfce) and crimson(#cc0000) ([oColor palette](https://github.com/Financial-Times/origami/blob/c47209e7f56f4a5d46437927ff73ca65012bbdfa/components/o-colors/src/scss/_palette.scss)). Is there any way to fine tune the colors?

GlynnPhillips: You can do shades/opacity of colors but the first thing I would do is speak with Mark and find out what those colors are. He should be using variants of ocolors so he should be aware of how to reproduce it

components/jsx/preferences-modal/index.js
ytcleon: I happen to forget that we can use CSS to hide the span when it is empty 🤦  Thanks for pointing it out.

components/jsx/preferences-modal/main.scss
```diff
@@ -47,14 +47,33 @@
...
 }
 
+// reference to the color/BG color of the error message
+// components/o-message/src/scss/_brand.scss#L21 in origami
 .n-myft-ui__preferences-modal-error {
 	display: block;
 	@include oTypographySans($scale: -1, $weight: 'regular', $style: 'normal');
-	color: oColorsByName('claret');
-	text-align: center;
+	color: oColorsMix('black', 'crimson', $percentage: 12.5);
```
GlynnPhillips, line 64: Thanks for checking these with Mark and making the updates

components/jsx/preferences-modal/index.js
```diff
@@ -3,10 +3,14 @@ import getToken from '../../../myft/ui/lib/get-csrf-token';
...
 
-const renderError = ({ message, preferencesModal }) => {
+const renderError = ({ message, preferencesModal, removeTopicButton }) => {
 	const errorElement = preferencesModal.querySelector('[data-component-id="myft-preference-modal-error"]');
 
 	errorElement.innerHTML = message;
+	// message is empty = remove existing error, re-enable remove button
+	if (removeTopicButton && message === '') {
+		removeTopicButton.removeAttribute('disabled');
+	}
```
GlynnPhillips, line 13: **Suggestion:** I am not sure this should live within this function because I find it a little confusing. I think resetting the disabled state should be done as part of the show/hide preference modal. I will leave a comment in that section to illustrate that

components/jsx/preferences-modal/index.js
```diff
@@ -94,21 +103,26 @@ const preferenceModalShowAndHide = ({ event, preferencesModal, conceptId }) => {
 		renderError({
 			message: '',
 			preferencesModal,
+			removeTopicButton,
 		});
```
GlynnPhillips, line 107: **Suggestion:** Somewhere here we could just do `removeTopicButton.removeAttribute('disabled');` regardless of whether its disabled or not (I don't think that will error)

components/jsx/preferences-modal/index.js
```diff
@@ -75,13 +79,18 @@ const tracking = (conceptId) => {
 	document.body.dispatchEvent(trackingEvent);
 };
 
-const preferenceModalHide = ({ event, preferencesModal }) => {
+const preferenceModalHide = ({ event, preferencesModal, removeTopicButton }) => {
```
GlynnPhillips, line 82: **Suggestion:** I know you didn't add this function but seeing your work has made me aware of it. I don't think this is needed as its own function. I think we can probably work it into the showAndHide function but we can do that later

components/jsx/preferences-modal/main.scss
```diff
@@ -47,14 +47,39 @@
...
+	transition: opacity 0.5s;
+
+	a {
+		@include oTypographySans($scale: -1, $weight: 'semibold', $style: 'normal');
+		color: oColorsMix('black', 'crimson', $percentage: 12.5);
+		text-decoration-color:  oColorsMix('black', 'crimson', $percentage: 12.5);
+	}
+}
+
+.n-myft-ui__preferences-modal-error:empty {
```
GlynnPhillips, line 81: **Praise:** Nice use of `:empty`

GlynnPhillips (approved):
Looks good, thanks for working on the feeback and working with Mark to improve the design & behaviour

PR #627, opened 2024-01-02
CON-3082: bump mocha version to v7

danieleruiz:
Description:
Mocha upgrade to v10 from v6 as part as toolkit migration steps
Ticket:
[CON-3082](https://financialtimes.atlassian.net/browse/CON-3082)](https://financialtimes.atlassian.net/browse/CON-3082)

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "karma-webpack": "^4.0.2",
     "lintspaces-cli": "^0.7.0",
     "lolex": "5.1.2",
-    "mocha": "6.2.2",
+    "mocha": "^7.2.0",
     "mockery": "2.1.0",
     "node-fetch": "2.6.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -63,15 +63,15 @@
     "karma-chrome-launcher": "3.2.0",
     "karma-firefox-launcher": "^1.0.0",
```
fenglish, line 64: These launchers were added before `karma-browserstack-launcher`. I wonder whether these two launchers are still necessary or not because it uses `karma-browserstack-launcher` 🤔

fenglish (commented):
Thank you for your change. It seems good to me. ~One suggestion. If these old launchers are not necessary anymore, please remove them~ 🙏

fenglish (approved):

PR #628, opened 2024-02-07
Disable eventListeners when saving related articles

MMSaleh-FT:
## Context

This PR fixes and issue that our team (professional-bolt-ons)  had when we added a save functionality for the related articles list in the next article page where the save event for saving a related article would trigger the same event listeners for the main article 
and display the UI component for adding article to the saved article list or notification that the article is unsaved 

### Before 

Screenshot from 2024-02-07 18-12-52
Screenshot from 2024-02-07 18-12-44

### After

Screenshot from 2024-02-07 18-16-20
Screenshot from 2024-02-07 18-16-16


myft/ui/lists.js
```diff
@@ -500,6 +500,14 @@ function openCreateListAndAddArticleOverlay (contentId, config) {
 function initialEventListeners () {
 
 	document.body.addEventListener('myft.user.saved.content.add', event => {
+		//Only show the create list overlay if the saved article is the main article, this is useful if there are other article listed in the article page
+		//For example having a related articles that can also be saved
+		const currentArticleId = window.location.pathname.split('/')[2];
```
jkerr321, line 505: Instead of using the window location here you can use the `data-content-id` value which is on the `<article>` element in the html (and other places) which is more robust than this solution. E.g. some articles will have vanity URLs like `ft.com/covid-tracker` in which case the `pathname` will not give you the id.

MMSaleh-FT: I should be able to access it

jkerr321 (approved):
Nice!

myft/ui/lists.js
```diff
@@ -500,6 +500,12 @@ function openCreateListAndAddArticleOverlay (contentId, config) {
 function initialEventListeners () {
 
 	document.body.addEventListener('myft.user.saved.content.add', event => {
+		//Only show the create list overlay if the saved article is the main article, this is useful if there are other article listed in the article page
+		//For example having a related articles that can also be saved
+		const currentContentId = document.querySelector('[data-content-id]').dataset.contentId;
+		if (currentContentId !== event.detail.subject) {
```
metart43, line 506: It's a bit risky way to access objects properties. Are we sure that both:
- `event.detail` & `document.querySelector('[data-content-id]').dataset` will never be `undefined`?

If they ever are this will cause an error

MMSaleh-FT: for `document.querySelector('[data-content-id]').dataset` it should never be `undefined` since I am getting the element that contains `data-*` selector which will be a part of the data set. you can check [https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) for more info

MMSaleh-FT: for `event.details` I think that you are right , will add null checks for it

myft/ui/lists.js
```diff
@@ -500,6 +500,12 @@ function openCreateListAndAddArticleOverlay (contentId, config) {
...
 
 	document.body.addEventListener('myft.user.saved.content.add', event => {
+		//Only show the create list overlay if the saved article is the main article, this is useful if there are other article listed in the article page
+		//For example having a related articles that can also be saved
+		const currentContentId = document.querySelector('[data-content-id]').dataset.contentId;
+		if (event && event.detail && currentContentId !== event.detail.subject) {
+			return;
+		}
 		event.stopPropagation();
 		const contentId = event.detail.subject;
```
metart43, line 510: non-blocking: I've noticed that `event.detail.subject;` is initialized on line:510. I'd just move the initialization up and reference this constant inside the if statement.

metart43 (approved):
It looks okay.

PR #640, opened 2024-04-29
CON-3317: fix aria label for alert icons

danieleruiz:
Issue: The state of the alert components was not clear for users of screen reading assistive
technologies
image

Solution:
change the aria-label text to fit with the current status of the instant alert of the topic

### DIFF
--- a/components/instant-alert/index.js
+++ b/components/instant-alert/index.js
@@ -27,7 +27,14 @@ function toggleButton (buttonEl, state) {
 
 	if (state !== isPressed) {
 		nextButtons.toggleState(buttonEl);
+		const name = buttonEl.getAttribute('data-name');
+		if (isPressed) {
+			buttonEl.setAttribute('aria-label', 'Instant alerts disabled for ' + name);
+		} else {
+			buttonEl.setAttribute('aria-label', 'Instant alerts enabled for ' + name);
+		}
 	}
+
 	buttonEl.removeAttribute('disabled');
 	buttonEl.setAttribute('value', !state);
 }
--- a/components/instant-alert/instant-alert.html
+++ b/components/instant-alert/instant-alert.html
@@ -13,10 +13,11 @@
 	<input type="hidden" value="http://www.ft.com/ontology/concept/Concept" name="directType">
 	{{/if}}
 	<button
-			aria-label="Get instant alerts for {{name}}"
 			{{#ifAll hasInstantAlert @root.cacheablePersonalisedUrl}}
+			aria-label="Instant alerts enabled for {{name}}"
 			aria-pressed="true"
 			{{else}}
+			aria-label="Instant alerts disabled for {{name}}"
 			aria-pressed="false"
 			{{/ifAll}}
 	class="n-myft-ui__button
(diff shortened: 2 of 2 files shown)


components/instant-alert/instant-alert.html
```diff
@@ -13,18 +13,19 @@
...
 			{{else}}
+			aria-label="Instant alerts disabled for {{name}}"
 			aria-pressed="false"
 			{{/ifAll}}
 	class="n-myft-ui__button
 	{{#variant}} n-myft-ui__button--{{this}}{{/variant}}
 	{{#size}} n-myft-ui__button--{{this}}{{/size}}
 	n-myft-ui__button--instant
 	n-myft-ui__button--instant-light"
-	data-alternate-label="Stop instant alerts for {{name}}"
```
gyss, line 27: why delete this prop? I understand we need `data-name` but probably we'll need to keep `data-alternate-label` (probably with different value depending on  `hasInstantAlert`)

components/instant-alert/index.js
```diff
@@ -27,7 +27,14 @@ function toggleButton (buttonEl, state) {
 
 	if (state !== isPressed) {
 		nextButtons.toggleState(buttonEl);
+		const name = buttonEl.getAttribute('data-name');
+		if (isPressed) {
+			buttonEl.setAttribute('aria-label', 'Instant alerts disabled for ' + name);
```
gyss, line 32: I was wondering, since this is the `aria-label` of a button, shouldn't the text express an **action**? something like "Disable instant alerts for xxx"

components/instant-alert/instant-alert.html
danieleruiz: data-alternate-label was a property that have a default behavior that is not working in this case. is like a property to give a button a alternate label but in this case we need to control the label depending on the current state.

components/instant-alert/index.js
danieleruiz: changed to a message that express the action and indicate the state

components/instant-alert/index.js
```diff
@@ -27,7 +27,14 @@ function toggleButton (buttonEl, state) {
 
 	if (state !== isPressed) {
 		nextButtons.toggleState(buttonEl);
+		const name = buttonEl.getAttribute('data-name');
+		if (isPressed) {
+			buttonEl.setAttribute('aria-label', 'Alerts ' + name + ' disabled, click to enable');
```
gyss, line 32: I'd discourage the use of the word "click" for screen readers: https://www.w3.org/QA/Tips/noClickHere.

Still thinking that something like `"Disable alerts for " + name` is the way to go. Anyways I'm no expert in accessibility. Maybe we can get advice from other engineer?

danieleruiz: i removed the word "click" to change, but this button have two has 2 functions: one to indicate the current status and the other one let user to change it
the message must indicate the current status and also the action you will do when you press it.
with this new message i think we give context to indicate both things

camillecroci (approved):
MIRAR BUENO PER ME (even though I prefer 'click'.Its not because you are visually impair that you are not clicking but I understand the link Jose shared so I wont fight. )

